PR #3, opened 2017-07-05
Remove some platform specific and auto-generated files

ad1217:
With the exception of Labrador.pro.user*, all of these files are auto-generated or backups. Labrador.pro.user* are user-specifc project files, so you probably want to keep your own copies.

You should also consider removing the compiled binaries (the bin directory) and just using Github's "releases" feature.
Additionally, you might want to split the PCB and code into separate repositories.


EspoTek:
You're right.  This was my first public repo and I included far more than necessary.  (I'm used to personal repos where I actually want to capture the exact state perfectly in case I ignore an important file.)

I think it might be best to just start a whole new repo, since there's almost 100MB of unneeded junk files in the old blobs.

ad1217:
That might be a good move, but I would suggest using [git filter-branch](https://git-scm.com/docs/git-filter-branch) to keep history. This allows you to remove files from history, as well as split directories off into other repos while preserving their history.

Personally, I would recommend at least 3 repositories: Desktop/Android program, AVR firmware, and PCB designs. (Also, possibly keep this one around but rename it for the Matlab interface and other old code)

EspoTek:
Looked up git filter-branch.  From what I understand, it will remove files from past revisions, as well, and the blobs will still be large?

PR #59, opened 2018-11-21
Altiumpcb

Ele4nor:
it's an altium based design but i think it's best to put it to another branch because the design is not compatible with the kicad design in current head.


EspoTek:
Thanks so much for this.  It looks great, but are you able to add the gerbers and .drl files (and maybe a .PDF of the schematic), so that others can use the designs without needing to aquire a copy of Altium?

EspoTek:
Thanks so much for this.  It's really cool.

PR #61, opened 2018-12-23
isoBufferBuffer refactor

SebastianMestre:
This improves the implementation and, arguably, the interface of isoBufferBuffer.

The main improvement comes from the reduction in memory allocation from a buffer 3 times as large as the maximum allowed query, to a buffer twice as large as the maximum query.

Also, the use of `char*` as a handle to the allocated buffer was replaced by `std::unique_ptr<char[]>`, removing a memory leak without the need to write a custom destructor.

The intent of this pull request is to simplify usage and make isoBufferBuffer look more like other C++ containers, including but not limited to exposing an iterator interface, and embracing const-correctness.

Examples of where this refactor would simplify usage are:

https://github.com/EspoTek/Labrador/blob/64b8b9311b14d500a4c9856f325f0e88aa9489f3/Desktop_Interface/i2cdecoder.cpp#L220-L221

Which, after this refactor, can become simply:

`console->setPlainText(QString::fromLocal8Bit(serialBuffer->begin(), serialBuffer->size())); `

And:

https://github.com/EspoTek/Labrador/blob/64b8b9311b14d500a4c9856f325f0e88aa9489f3/Desktop_Interface/i2cdecoder.cpp#L35-L36

Which can become

`serialBuffer->clear();`

If this pull request is merged, i plan on moving on to update the points where this is used to use the new interface, and later remove the old interface.

NOTE: this pull request uses C++14 features. If required, it can be modified to only use features of up to C++11.

### DIFF
--- a/Desktop_Interface/isobufferbuffer.cpp
+++ b/Desktop_Interface/isobufferbuffer.cpp
@@ -20,32 +20,64 @@
  * half of the allocated buffer, which is a notable improvement.
  */
 
-// TODO: go through the usages of this class and adjust the
-// size of the requested buffer to accomodate the improved memory
-// efficiency of the new algorithm
 
-isoBufferBuffer::isoBufferBuffer(
+// TODO: go through the usages of this class and:
+// 1. adapt code to use the new interface and remove the old one
+// 2. adjust the size of the requested buffer to accomodate
+// the improved memory efficiency of the new algorithm
+
+isoBufferBuffer::isoBufferBuffer (
 	uint32_t length
 ) :
-	data_ ( std::make_unique<char[]>(length*2) ),
-	capacity_ (length)
+	data_(std::make_unique<char[]>(length*2)),
+	capacity_(length)
 {}
 
-void isoBufferBuffer::insert (
-	char c
-) {
+// Adds a character to the end of the buffer
+void isoBufferBuffer::insert ( char c ) {
+	// Add character to first half of the buffer
 	data_[top_] = c;
+	// Then to the second
 	data_[top_+capacity_] = c;
 
-	top_ = (top_+1) % capacity_;
-	size_ = std::min(size_+1, capacity_);
+	// Loop the buffer index if necessary and update size accordingly
+	top_ = (top_ + 1) % capacity_;
+	size_ = std::min(size_ + 1, capacity_);
 }
 
-char const * isoBufferBuffer::query (
-	uint32_t length
-) const {
-	uint32_t offset = top_ < length ? capacity_ : 0 ;
-	return data_.get() + top_ - length + offset;
+void isoBufferBuffer::insert ( char const * s ) {
+	while ( *s )
+		insert(*s++);
+}
+
+void isoBufferBuffer::insert ( std::string const & s ) {
+	insert(s.c_str());
+}
+
+char const* isoBufferBuffer::query ( uint32_t count ) const {
+	if ( count > capacity_ )
+		qFatal("isoBufferBuffer::query : you may not request more items than the capacity of the buffer");
+
+	if ( count > size_ )
+		qFatal("isoBufferBuffer::query : you may not request more items than inserted");
+
+	// we offset the returned pointer to point into the second half of the buffer
+	auto offset = capacity_;
+	char const * ptr = data_.get() + top_ - count + offset;
+	return ptr;
+}
+
+void isoBufferBuffer::clear () {
+	top_ = 0;
+	size_ = 0;
+};
+
+char const * isoBufferBuffer::begin () const {
+	return data_.get() + top_ - size_ + capacity_;
+}
+
+char const * isoBufferBuffer::end () const {
+	return data_.get() + top_ + capacity_;
 }
 
 uint32_t isoBufferBuffer::size () const {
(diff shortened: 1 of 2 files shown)


SebastianMestre:
NOTE1: AppVeyor build failed to fo authentication failure.
NOTE2: Travis CI build failed due to usage of C++14 features.

EspoTek:
Hey, you the guy who started the thread on Reddit?
Thanks for the PR.  I'm sure you were just thinking of it in terms of an intellectual exercise, but this patch will actually improve performance on low-memory systems such as the Raspberry Pi.

Had a quick look and it mostly looks good.  I'll need to go over it properly after Christmas to give full feedback, though.  Just dead tired right now.  Thanks once again, though.  I'll be back soon with a full review!  :)

SebastianMestre:
Yup, I'm that guy from Reddit.

Haha, that's great. I didn't have that particularly in mind but I *was* trying to improve things as much as i possibly could.

Ah, good to hear! Happy holidays!

Desktop_Interface/isobufferbuffer.cpp
```diff
@@ -1,49 +1,114 @@
 #include "isobufferbuffer.h"
 
-isoBufferBuffer::isoBufferBuffer(uint32_t length)
-    : bufferLength(length)
-{
-    mid = bufferLength/2;
-    buffer = (char *) malloc((bufferLength * 3) / 2);
+/* isoBufferBuffer is implemented as two consecutive, duplicate,
```
EspoTek: Good idea leaving a comment like this. Makes understanding everything much easier when you haven't looked at it in 2 years. Definitely should have written something like this at the time.

Desktop_Interface/isobufferbuffer.cpp
```diff
@@ -1,49 +1,114 @@
...
+
+isoBufferBuffer::isoBufferBuffer (
+	uint32_t length
+) :
+	data_(std::make_unique<char[]>(length*2)),
+	capacity_(length)
+{}
+
+// Adds a character to the end of the buffer
+void isoBufferBuffer::insert ( char c ) {
```
EspoTek: This function is much neater now.  Comments are clear and flow makes more sense.  Literally can't offer a single recommendation to improve it.

Desktop_Interface/isobufferbuffer.cpp
```diff
@@ -1,49 +1,114 @@
...
+	// Then to the second
+	data_[top_+capacity_] = c;
+
+	// Loop the buffer index if necessary and update size accordingly
+	top_ = (top_ + 1) % capacity_;
+	size_ = std::min(size_ + 1, capacity_);
+}
+
+void isoBufferBuffer::insert ( char const * s ) {
+	while ( *s )
```
EspoTek: It's generally better to be explicit about comparisons, rather than relying on the fact that specific constructs (e.g. nullptr, null terminator on string) is zero.
It won't affect behaviour on any remotely decent compiler, but it makes your intentions clearer to other people reading the code.
In this case, a better way of writing it would be "while(*s != '\0')".

Desktop_Interface/isobufferbuffer.h
```diff
@@ -1,27 +1,59 @@
...
-class isoBufferBuffer
-{
+/** @file isobufferbuffer.h
+ *  @brief This  module  implements  a  data structure that allows
+ *  insertion  of  single  characters  and  a  view  of the last N
+ *  inserted characters in constant time.
+ *
+ *  To  obtain  such  complexity,  a  double  ring buffer is used.
+ *  That  is,  two  identical  ring buffers are mantained adjacent
+ *  in  memory.  If we always return a pointer to the begging of a
```
EspoTek: begging => beginning

Desktop_Interface/isobufferbuffer.h
```diff
@@ -1,27 +1,59 @@
...
+ *  range  that ends in the second buffer, we will always return a
+ *  valid  address(*),  even  when the requested length is greater
+ *  than  the  current position being inserted into in the buffer.
+ *
+ *  (*) By  valid  address  I  mean  that  both the addresses that
+ *  represent  the  begining and end of the requested query result
```
EspoTek: begining => beginning

Desktop_Interface/isobufferbuffer.h
```diff
@@ -1,27 +1,59 @@
...
+ */
+class isoBufferBuffer {
 public:
-    isoBufferBuffer(uint32_t length);
-    void add(uint8_t newByte);
-    void add(char newChar);
-    void add(std::string newString);
-    char *get(uint32_t length);
-    uint32_t getNumCharsInBuffer();
+	isoBufferBuffer ( uint32_t length );
```
EspoTek: Don't know why you've added the double tabs here.

Desktop_Interface/isobufferbuffer.cpp
```diff
@@ -1,49 +1,114 @@
...
+
+// TODO: go through the usages of this class and:
+// 1. adapt code to use the new interface and remove the old one
+// 2. adjust the size of the requested buffer to accomodate
+// the improved memory efficiency of the new algorithm
+
+isoBufferBuffer::isoBufferBuffer (
+	uint32_t length
+) :
+	data_(std::make_unique<char[]>(length*2)),
```
EspoTek: Good call with the unique_ptr.

Desktop_Interface/isobufferbuffer.h
SebastianMestre: Ah, that's weird. It doesn't look like that for me even though i use 4 space tabs.

EspoTek:
Oh, and Happy Holidays!

Desktop_Interface/isobufferbuffer.h
SebastianMestre: Oops, I always have trouble spelling that word. Coding late at night doesn't help either.

EspoTek:
One more thing, see the following commit:
https://github.com/EspoTek/Labrador/commit/14a5ab67ffa391856737cc7d133ef9a774e0de0f

The .pro file is parsed by `qmake` and used to generate nmake/makefiles depending on the platform.  If you need anything in particular you can usually just add a line there.  Don't worry too much about neatness.  :)

EspoTek:
Thanks for this, PR looks great now.
I've edited the travis config file so it should build properly with C++14 (it uses gcc-7 now) and I'm merging it into the main branch.  :)

If you plan on doing more PRs, please let me know and I'll make sure to get you some hardware!

Thanks once again,
~Chris

SebastianMestre:
> Thanks for this, PR looks great now.
> I've edited the travis config file so it should build properly with C++14 (it uses gcc-7 now) and I'm merging it into the main branch. :)
> 
> If you plan on doing more PRs, please let me know and I'll make sure to get you some hardware!
> 
> Thanks once again,
> ~Chris

About that hardware, I would love to get a piece. I'm not based in the US though, would that be an issue?

PR #64, opened 2019-01-14
Isobuffer refactor

SebastianMestre:
I had a go at refactoring isoBuffer. The main benefit is that i managed to reduce duplication, and that i managed to make the logic in some of these functions simpler.

I also added some comments which will hopefully make isoBuffer easier to mantain in the future.

There are a few caveats though:
~First, I did not go over the isoBuffer::cap_xnfromLast family of functions but i believe that these can be simplified and DRYed up quite easily.~
And second, i went pretty hard on isoBuffer::readBuffer, and it ended up not resembling the original function much or at all. I fear it no longer does the same thing it used to do... I tested it a bit and i think it's fine but i can't tell for sure. I added a comment explaining what i think this function is meant to do, and my implementation seems to do what i want (that is, it passes my tests). But I'm still quite unsure.

EDIT: I have gone through and further removed duplication by extracting the common functionality of cap_x0fromLast, cap_x1fromLast, and cap_x2fromLast to a separate function, capSample. Since this function shouldn't really be accesible from anywhere else, it is declared and defined in isoBuffer.cpp

### DIFF
--- a/Desktop_Interface/isobuffer.cpp
+++ b/Desktop_Interface/isobuffer.cpp
@@ -203,23 +203,23 @@ void isoBuffer::glitchInsert(short type)
 void isoBuffer::enableFileIO(QFile* file, int samplesToAverage, qulonglong max_file_size)
 {
 
-	//Open the file
+	// Open the file
 	file->open(QIODevice::WriteOnly);
 	currentFile = file;
 
-	//Add the header
+	// Add the header
 	char headerLine[256];
 	sprintf(headerLine, "EspoTek Labrador DAQ V1.0 Output File\nAveraging = %d\nMode = %d\n", samplesToAverage, virtualParent->driver->deviceMode);
 	currentFile->write(headerLine);
 
-	//Set up the isoBuffer for DAQ
+	// Set up the isoBuffer for DAQ
 	fileIO_maxIncrementedSampleValue = samplesToAverage;
 	fileIO_max_file_size = max_file_size;
 	fileIO_sampleCount = 0;
 	fileIO_numBytesWritten = 0;
 	average_sample_temp = 0;
 
-	//Enable DAQ
+	// Enable DAQ
 	fileIOEnabled = true;
 
 	qDebug("File IO enabled, averaging %d samples, max file size %lluMB", samplesToAverage, max_file_size/1000000);
@@ -309,7 +309,7 @@ int capSample(isoBuffer const& self, int rbegin, int target, double seconds, dou
 	return -1;
 }
 
-//For capacitance measurement. x0, x1 and x2 are all various time points used to find the RC coefficient.
+// For capacitance measurement. x0, x1 and x2 are all various time points used to find the RC coefficient.
 int isoBuffer::cap_x0fromLast(double seconds, double vbot)
 {
 	return capSample(*this, 0, NUM_SAMPLES_SEEKING_CAP, seconds, vbot, X0_COMP_FTOR);
@@ -335,6 +335,9 @@ void isoBuffer::serialManage(double baudRate, int type, UartParity parity)
 	if (decoder == NULL)
 	{
 		decoder = new uartStyleDecoder(this);
+		// TODO: Look into using the type safe version of connect.
+		// NOTE: I believe Qt has a type-safe version of this, without the macros and the
+		// explicit signature and stuff, i think it uses member-function pointers instead.
 		connect(decoder, SIGNAL(wireDisconnected(int)), virtualParent, SLOT(serialNeedsDisabling(int)));
 	}
 	if (stopDecoding)
--- a/Desktop_Interface/isobuffer.cpp
+++ b/Desktop_Interface/isobuffer.cpp
@@ -40,7 +40,8 @@ void isoBuffer::insertIntoBuffer(short item)
 	buffer[back] = item;
 	back++;
 	insertedCount++;
-	if(insertedCount > bufferEnd) {
+	if (insertedCount > bufferEnd)
+	{
 		insertedCount = bufferEnd+1;
 	}
 	if (back > bufferEnd)
@@ -171,26 +172,21 @@ short* isoBuffer::readBuffer(double sampleWindow, int numSamples, bool singleBit
 	free(readData);
 	readData = (short*) calloc(numSamples, sizeof(short));
 
-	// NOTE: this min seems unnecesary.
-	double pos = std::min(0, back - delaySamples - 1);
+	double itr = delaySamples + 1;
 	for (int i = 0; i < numSamples; i++)
 	{
-		while (pos < 0)
-			pos += bufferEnd;
+		while (itr > insertedCount)
+			itr -= insertedCount;
 
-		int idx = pos;
+		readData[i] = bufferAt(int(itr));
 
 		if (singleBit)
 		{
-			int subIdx = 8*(pos - floor(pos));
-			readData[i] = buffer[idx] & (1 << subIdx);
-		}
-		else
-		{
-			readData[i] = buffer[idx];
+			int subIdx = 8*(-itr-floor(-itr));
+			readData[i] &= (1 << subIdx);
 		}
 
-		pos -= timeBetweenSamples;
+		itr += timeBetweenSamples;
 	}
 
 	return readData;
--- a/Desktop_Interface/isobuffer.cpp
+++ b/Desktop_Interface/isobuffer.cpp
@@ -42,7 +42,7 @@ void isoBuffer::insertIntoBuffer(short item)
 	}
 }
 
-short isoBuffer::bufferAt(int idx)
+short isoBuffer::bufferAt(int idx) const
 {
 	return buffer[back - idx];
 }
@@ -245,7 +245,7 @@ void isoBuffer::disableFileIO()
 	return;
 }
 
-double isoBuffer::sampleConvert(short sample, int TOP, bool AC)
+double isoBuffer::sampleConvert(short sample, int TOP, bool AC) const
 {
 
 	double scope_gain = (double)(virtualParent->driver->scopeGain);
@@ -263,7 +263,7 @@ double isoBuffer::sampleConvert(short sample, int TOP, bool AC)
 	return voltageLevel;
 }
 
-short isoBuffer::inverseSampleConvert(double voltageLevel, int TOP, bool AC)
+short isoBuffer::inverseSampleConvert(double voltageLevel, int TOP, bool AC) const
 {
 	double scope_gain = (double)(virtualParent->driver->scopeGain);
 	short sample;
--- a/Desktop_Interface/isobuffer.cpp
+++ b/Desktop_Interface/isobuffer.cpp
@@ -148,12 +148,13 @@ short* isoBuffer::readBuffer(double sampleWindow, int numSamples, bool singleBit
 
 	free(readData);
 
+	readData = (short*) calloc(numSamples, sizeof(short));
+
+	// TODO: replace by return nullptr and add error handling upstream
 	if(delaySamples+1 > insertedCount)
 	{
-		return readData = nullptr;
+		return readData;
 	}
-
-	readData = (short*) calloc(numSamples, sizeof(short));
 
 	double itr = delaySamples + 1;
 	for (int i = 0; i < numSamples; i++)
--- a/Desktop_Interface/isobuffer.cpp
+++ b/Desktop_Interface/isobuffer.cpp
@@ -198,11 +198,6 @@ void isoBuffer::gainBuffer(int gain_log)
 	}
 }
 
-// NOTE: glitch insert is a no-op
-void isoBuffer::glitchInsert(short type)
-{
-}
-
 void isoBuffer::enableFileIO(QFile* file, int samplesToAverage, qulonglong max_file_size)
 {
 
--- a/Desktop_Interface/isobuffer.cpp
+++ b/Desktop_Interface/isobuffer.cpp
@@ -48,14 +48,14 @@ short isoBuffer::bufferAt(int idx) const
 bool isoBuffer::maybeOutputSampleToFile(double convertedSample)
 {
 	/*
-	 * This function adds a sample to an accumulator and bumps the sample count
-	 * after the sample count hits some threshold, that sample is outputted to a
-	 * file. If this 'saturates' the file, then fileIO is disabled.
+	 * This function adds a sample to an accumulator and bumps the sample count.
+	 * After the sample count hits some threshold, the accumulated sample is
+	 * outputted to a file. If this 'saturates' the file, then fileIO is disabled.
 	 */
 	average_sample_temp += convertedSample;
 	fileIO_sampleCount++;
 
-	//Check to see if we can write a new sample to file
+	// Check to see if we can write a new sample to file
 	if (fileIO_sampleCount == fileIO_maxIncrementedSampleValue)
 	{
 		char numStr[32];
@@ -68,17 +68,17 @@ bool isoBuffer::maybeOutputSampleToFile(double convertedSample)
 			currentColumn = 0;
 		}
 
-		//Reset the average and sample count for next data point
+		// Reset the average and sample count for next data point
 		fileIO_sampleCount = 0;
 		average_sample_temp = 0;
 
-		//Check to see if we've reached the max file size.
-		if (fileIO_max_file_size != 0) //value of 0 means "no limit"
+		// Check to see if we've reached the max file size.
+		if (fileIO_max_file_size != 0) // value of 0 means "no limit"
 		{
-			fileIO_numBytesWritten += 9;  //7 chars for the number, 1 for the comma and 1 for the space = 9 bytes per sample.
+			fileIO_numBytesWritten += 9;  // 7 chars for the number, 1 for the comma and 1 for the space = 9 bytes per sample.
 			if (fileIO_numBytesWritten >= fileIO_max_file_size)
 			{
-				fileIOEnabled = false; //Just in case signalling fails.
+				fileIOEnabled = false; // Just in case signalling fails.
 				fileIOinternalDisable();
 				return false;
 			}
--- a/Desktop_Interface/isobuffer.cpp
+++ b/Desktop_Interface/isobuffer.cpp
@@ -39,7 +39,6 @@ void isoBuffer::insertIntoBuffer(short item)
 	if (back > bufferEnd)
 	{
 		back = 0;
-		firstTime = false;
 	}
 }
 
@@ -192,7 +191,6 @@ void isoBuffer::clearBuffer()
     }
 
     back = 0;
-    firstTime = true;
 }
 
 void isoBuffer::gainBuffer(int gain_log)
--- a/Desktop_Interface/isobuffer.cpp
+++ b/Desktop_Interface/isobuffer.cpp
@@ -2,7 +2,8 @@
 #include "isodriver.h"
 #include "uartstyledecoder.h"
 
-isoBuffer::isoBuffer(QWidget *parent, int bufferLen, isoDriver *caller, unsigned char channel_value) : QWidget(parent)
+isoBuffer::isoBuffer(QWidget *parent, int bufferLen, isoDriver *caller, unsigned char channel_value)
+	: QWidget(parent)
 {
     buffer = (short *) calloc(bufferLen*2, sizeof(short));
     bufferEnd = bufferLen-1;
@@ -12,6 +13,7 @@ isoBuffer::isoBuffer(QWidget *parent, int bufferLen, isoDriver *caller, unsigned
     virtualParent = caller;
     channel = channel_value;
 }
+
 void isoBuffer::openFile(QString newFile)
 {
     if (fptr != NULL){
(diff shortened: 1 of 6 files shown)


EspoTek:
Thanks for another review.  I'll sit down and do proper reviews + testing when I've got a bit of time.  In the meantime, I'll just say that the function name `maybeOutputSampleToFile` is absolutely brilliant.

SebastianMestre:
Got it! Thank you!

SebastianMestre:
I have plans for further refactorings but I think this is already too much for a single PR. I know reviewing code can be time consuming, sorry about that.

Desktop_Interface/isobuffer.h
```diff
@@ -18,62 +19,90 @@ class isoDriver;
...
 enum class UartParity : uint8_t;
 
-//isoBuffer is a generic class that enables O(1) read times (!!!) on all read/write operations, while maintaining a huge buffer size.
-//Imagine it as a circular buffer, but with access functions specifically designed for isochronous data from an Xmega.
+// isoBuffer is a generic class that enables O(1) read times (!!!) on all
+// read/write operations, while maintaining a huge buffer size.
+// Imagine it as a circular buffer, but with access functions specifically
+// designed for isochronous data from an Xmega.
 
 #define CONSOLE_UPDATE_TIMER_PERIOD (ISO_PACKETS_PER_CTX * 4)
```
EspoTek: If possible, changing a lot of these #defines to globals (or variables owned by a specific object and signals sent when the value changes) would be really handy as it would allow the value of these variables to change at runtime.  This would be especially handy for many of the variables in the USB driver as they control frame rates.

Desktop_Interface/isobuffer.cpp
```diff
@@ -2,391 +2,364 @@
...
-    samplesPerSecond = (double) bufferLen/(double)21;
-    samplesPerSecond = samplesPerSecond/375*VALID_DATA_PER_375;
-    sampleRate_bit = samplesPerSecond * 8;
-    virtualParent = caller;
-    channel = channel_value;
+	buffer = (short*) calloc(bufferLen*2, sizeof(short));
+	bufferEnd = bufferLen-1;
+	samplesPerSecond = (double) bufferLen/(double)21;
+	samplesPerSecond = samplesPerSecond/375*VALID_DATA_PER_375;
+	sampleRate_bit = samplesPerSecond * 8;
```
EspoTek: Many of the assignments in the constructor belong in the initialiser list.

Desktop_Interface/isobuffer.cpp
```diff
@@ -2,391 +2,364 @@
 #include "isodriver.h"
 #include "uartstyledecoder.h"
 
-isoBuffer::isoBuffer(QWidget *parent, int bufferLen, isoDriver *caller, unsigned char channel_value) : QWidget(parent)
```
EspoTek: Here and elsewhere:
It's safer to use the cstdint types such as uint8_t and int16_t, especially in a project such as this where bit positions matter.

Desktop_Interface/isobuffer.cpp
```diff
@@ -2,391 +2,364 @@
...
+void isoBuffer::insertIntoBuffer(short item)
+{
+	buffer[back] = item;
+	back++;
+	insertedCount++;
+	if (insertedCount > bufferEnd)
+	{
+		insertedCount = bufferEnd+1;
+	}
+	if (back > bufferEnd)
```
EspoTek: As above: it's best to separate independent if() statements.  Keeping them together can make them look like a single if-else construct at a glance.

Desktop_Interface/isobuffer.cpp
```diff
@@ -2,391 +2,364 @@
...
+	const double timeBetweenSamples = sampleWindow * samplesPerSecond / numSamples;
+	const int delaySamples = delayOffset * samplesPerSecond;
+
+	free(readData);
+	readData = (short*) calloc(numSamples, sizeof(short));
+
+	double itr = delaySamples + 1;
+	for (int i = 0; i < numSamples; i++)
+	{
+		while (itr > insertedCount)
```
EspoTek: I'm not quite sure that this is the best behaviour.  It looks like it's providing more recent data for the case when the delay is larger than the valid buffer length.  In this case, we probably just want to return nullptr (and check for errors upstream).

Desktop_Interface/isobuffer.cpp
```diff
@@ -2,391 +2,364 @@
...
+				fileIOEnabled = false; //Just in case signalling fails.
+				fileIOinternalDisable();
+				return false;
+			}
+		}
+	}
+	return true;
 }
 
 void isoBuffer::writeBuffer_char(char* data, int len)
```
EspoTek: Wow.  Good work refactoring these.  Just a little bit more concise now!
Could template use allow us to roll them into a single function now, though, since they share exactly the same source code?

Desktop_Interface/isobuffer.cpp
```diff
@@ -2,391 +2,364 @@
...
+{
+	/*
+	 * This function adds a sample to an accumulator and bumps the sample count
+	 * after the sample count hits some threshold, that sample is outputted to a
+	 * file. If this 'saturates' the file, then fileIO is disabled.
+	 */
+	average_sample_temp += convertedSample;
+	fileIO_sampleCount++;
+
+	//Check to see if we can write a new sample to file
```
EspoTek: Here and elsewhere:
space after //

Desktop_Interface/isobuffer.cpp
```diff
@@ -2,391 +2,364 @@
...
+
+short isoBuffer::bufferAt(int idx) const
+{
+	return buffer[back - idx];
+}
+
+bool isoBuffer::maybeOutputSampleToFile(double convertedSample)
+{
+	/*
+	 * This function adds a sample to an accumulator and bumps the sample count
```
EspoTek: The accumulator is used to take sample averages.  For example, in the case where the user wants to output one data sample to a file per second but the scope reads at 750ksps, this will average 750,000 samples and then write the average to the file.
However, there can be a long time and multiple functions calls before it actually gets that many samples, hence the need for the long-term accumulator.[

Desktop_Interface/isobuffer.cpp
```diff
@@ -2,391 +2,364 @@
...
+			bool keepOutputting = maybeOutputSampleToFile(convertedSample);
+
+			if (!keepOutputting) break;
+		}
+	}
 }
 
-
-short *isoBuffer::readBuffer(double sampleWindow, int numSamples, bool singleBit, double delayOffset)
+short* isoBuffer::readBuffer(double sampleWindow, int numSamples, bool singleBit, double delayOffset)
```
EspoTek: Wow.  Your implementation is more than a bit cleaner.
Even with that giant comment it's half the size!

Desktop_Interface/isobuffer.cpp
```diff
@@ -2,391 +2,364 @@
...
+		if (singleBit)
+		{
+			int subIdx = 8*(-itr-floor(-itr));
+			readData[i] &= (1 << subIdx);
+		}
+
+		itr += timeBetweenSamples;
+	}
+
+	return readData;
```
EspoTek: I'm not sure that returning a plain pointer is the best way to go here, as it means that the caller must manually free it (and who owns the pointer is not always clear!).  Since we have C++11, std::unique_ptr and std::move could solve this problem.

Desktop_Interface/isobuffer.cpp
```diff
@@ -2,391 +2,364 @@
...
+	for (int i=0; i<bufferEnd; i++)
+	{
+		if (gain_log == -1) buffer[i] *= 2;
+		else buffer[i] /= 2;
+	}
 }
 
-
+// NOTE: glitch insert is a no-op
 void isoBuffer::glitchInsert(short type)
```
EspoTek: This was originally intended to store metadata in the top bits of the samples, as only 12 or 8 bit samples are stored in there.  Of course, this never ended up being implemented.  :(

Desktop_Interface/isobuffer.cpp
```diff
@@ -2,391 +2,364 @@
...
+	file->open(QIODevice::WriteOnly);
+	currentFile = file;
 
-    //Add the header
-    char headerLine[256];
-    sprintf(headerLine, "EspoTek Labrador DAQ V1.0 Output File\nAveraging = %d\nMode = %d\n", samplesToAverage, virtualParent->driver->deviceMode);
-    currentFile->write(headerLine);
+	//Add the header
+	char headerLine[256];
+	sprintf(headerLine, "EspoTek Labrador DAQ V1.0 Output File\nAveraging = %d\nMode = %d\n", samplesToAverage, virtualParent->driver->deviceMode);
```
EspoTek: Putting this in a constant string in an anonymous namespace at the top of the .cpp would be a good idea, as otherwise it's easy to forget to change the string when the algorithm is updated.

Desktop_Interface/isobuffer.cpp
```diff
@@ -2,391 +2,364 @@
...
+	double voltageLevel;
 
-    voltageLevel = (sample * (vcc/2)) / (frontendGain*scope_gain*TOP);
-    if (virtualParent->driver->deviceMode != 7) voltageLevel += voltage_ref;
-    #ifdef INVERT_MM
-        if(virtualParent->driver->deviceMode == 7) voltageLevel *= -1;
-    #endif
+	voltageLevel = (sample * (vcc/2)) / (frontendGain*scope_gain*TOP);
+	if (virtualParent->driver->deviceMode != 7) voltageLevel += voltage_ref;
+#ifdef INVERT_MM
```
EspoTek: Haha, I'm sorry for this #define.  Probably threw you a bit.  There was a particular old build (possibly even on a single board prototype) where the multimeter samples had their sign inverted by mistake.

Desktop_Interface/isobuffer.cpp
```diff
@@ -2,391 +2,364 @@
...
 
-    //voltageLevel = (sample * (vcc/2)) / (frontendGain*scope_gain*TOP);
-    sample = (voltageLevel * (frontendGain*scope_gain*TOP))/(vcc/2);
-    return sample;
+	//voltageLevel = (sample * (vcc/2)) / (frontendGain*scope_gain*TOP);
+	sample = (voltageLevel * (frontendGain*scope_gain*TOP))/(vcc/2);
+	return sample;
 }
 
 #define NUM_SAMPLES_SEEKING_CAP (20)
```
EspoTek: This would be easier to find in an anonymous namespace at the top of the cpp.

Desktop_Interface/isobuffer.cpp
```diff
@@ -2,391 +2,364 @@
...
 
 #ifdef INVERT_MM
-    #define X0_COMPARISON_CAP >
-    #define X1_X2_COMPARISON_CAP <
+constexpr auto X0_COMP_FTOR = std::greater<int> {};
+constexpr auto X1_X2_COMP_FTOR = std::less<int> {};
 #else
-    #define X0_COMPARISON_CAP <
-    #define X1_X2_COMPARISON_CAP >
+constexpr auto X0_COMP_FTOR = std::less<int> {};
```
EspoTek: Thank you for changing this.  I almost threw up when I saw the #define being used for an operator!

Desktop_Interface/isobuffer.cpp
```diff
@@ -2,391 +2,364 @@
...
-        if (num_found > NUM_SAMPLES_SEEKING_CAP){
-            return samplesInPast-i + x0;
-        }
-
-    }
-    return -1;
-}
+	if (self.back < samples + rbegin) return -1;
+
+	short sample = self.inverseSampleConvert(value, 2048, 0);
```
EspoTek: This "2048" value should probably be a defined constant in an anonymous namespace at the top of the .cpp.
It represents the top value of the multimeter ADC, so a name like "kTopMultimeter" would be good.  Also, you should be able to constexpr it.

Desktop_Interface/isobuffer.cpp
```diff
@@ -2,391 +2,364 @@
...
-        if(num_found < 0){
-            num_found = 0;
-        }
-        if (num_found > NUM_SAMPLES_SEEKING_CAP){
-            return samplesInPast-i;
-        }
-    }
-    return -1;
-}
+int capSample(isoBuffer const& self, int rbegin, int target, double seconds, double value, auto comp)
```
EspoTek: Why do you need the const ref to yourself?

EspoTek: (That being said though, just read the function and damn is this one more efficient)

Desktop_Interface/isobuffer.cpp
```diff
@@ -2,391 +2,364 @@
...
+{
+	return capSample(*this, -x0, NUM_SAMPLES_SEEKING_CAP, seconds, vbot, X1_X2_COMP_FTOR);
+}
 
+int isoBuffer::cap_x2fromLast(double seconds, int x1, double vtop)
+{
+	return capSample(*this, -x1, NUM_SAMPLES_SEEKING_CAP, seconds, vtop, X1_X2_COMP_FTOR);
+}
 
+// NOTE: type appears to be unused
```
EspoTek: Yep, that's correct.  Type is unused and should be removed.
This function was originally going to manage i2c as well as rs-232 but that did not end up being the case.

Desktop_Interface/isobuffer.h
```diff
@@ -18,62 +19,90 @@ class isoDriver;
...
+	bool maybeOutputSampleToFile(double convertedSample);
+	double sampleConvert(short sample, int TOP, bool AC) const;
+	short inverseSampleConvert(double voltageLevel, int TOP, bool AC) const;
+
+	int cap_x0fromLast(double seconds, double vbot);
+	int cap_x1fromLast(double seconds, int x0, double vbot);
+	int cap_x2fromLast(double seconds, int x1, double vtop);
+	void serialManage(double baudRate, int type, UartParity parity);
+
+// ---- MEMBER VARIABLES ----
```
EspoTek: Prefixing these with `m_` would be good for consistency when the code is finally refactored!

Desktop_Interface/isobuffer.cpp
SebastianMestre: That's because i made that function be a free function instead of a private method.

No, reason, really. It was just late at night and made a weird decision.
It'll be a private member function on next push.

Desktop_Interface/isobuffer.h
SebastianMestre: Ah, yes. Will do.

Desktop_Interface/isobuffer.cpp
SebastianMestre: Ah, got it.

Desktop_Interface/isobuffer.cpp
SebastianMestre: Oh yes. I was thinking about that but was unsure about what your policy on templates was, so I didn't want to try that just yet.

Desktop_Interface/isobuffer.cpp
SebastianMestre: Oh, I see. I'll fix that then. I might have some questions later, about what would be a good way to handle nulls upstream, if that's ok.

Desktop_Interface/isobuffer.cpp
SebastianMestre: oops... I didn't notice i did that. I'll sort that out.

Desktop_Interface/isobuffer.cpp
```diff
@@ -2,391 +2,364 @@
...
+	buffer = (short*) calloc(bufferLen*2, sizeof(short));
+	bufferEnd = bufferLen-1;
+	samplesPerSecond = (double) bufferLen/(double)21;
+	samplesPerSecond = samplesPerSecond/375*VALID_DATA_PER_375;
+	sampleRate_bit = samplesPerSecond * 8;
+	virtualParent = caller;
+	channel = channel_value;
 }
+
 void isoBuffer::openFile(QString newFile)
```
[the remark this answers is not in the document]
SebastianMestre: I tried removing it and everything still builds, so it seems it was never called, too.

Desktop_Interface/isobuffer.h
SebastianMestre: I am not super familiar with Qt's signal-slot model, but I'll look into it.

SebastianMestre: I can totally change it to a global for now though.

Desktop_Interface/isobuffer.h
```diff
@@ -18,62 +19,90 @@ class isoDriver;
...
+
+//	Presentantion?
+// NOTE: it seems like these are never initialized but they are used as though they were...
```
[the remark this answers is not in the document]
SebastianMestre: Ah, i see. i grepped for it and didn't find it, must've missed it.

Desktop_Interface/isobuffer.cpp
SebastianMestre: Also: I happened to look through isoBuffer_file and it looks like it should eventually be possible to extract the common functionality of isoBuffer and isoBuffer_file to a separate class :)

Desktop_Interface/isobuffer.cpp
SebastianMestre: done

Desktop_Interface/isobuffer.cpp
```diff
@@ -2,391 +2,364 @@
...
+	: QWidget(parent)
 {
-    buffer = (short *) calloc(bufferLen*2, sizeof(short));
-    bufferEnd = bufferLen-1;
-    samplesPerSecond = (double) bufferLen/(double)21;
-    samplesPerSecond = samplesPerSecond/375*VALID_DATA_PER_375;
-    sampleRate_bit = samplesPerSecond * 8;
-    virtualParent = caller;
-    channel = channel_value;
+	buffer = (short*) calloc(bufferLen*2, sizeof(short));
```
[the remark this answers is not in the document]
SebastianMestre: Got it! will definitely do that for the next PR

Desktop_Interface/isobuffer.cpp
SebastianMestre: I removed it now

Desktop_Interface/isobuffer.cpp
SebastianMestre: Ah yes. I'll take some modifying of caller code though, so I'll put it off till the next PR

Desktop_Interface/isobuffer.cpp
SebastianMestre: I decided to remove it for now, since it wasn't necesary. I would love to implement that feature in the future if you want.

SebastianMestre:
All right, i think now it is in a state more suitable for merging. I'm really sorry about this huge PR, I faced the refactoring with a target in mind and just started going at it without realizing how much I had changed and how much I needed to change.

I'm still not done but this feels like a good place to stop for now. I have a sort of ambitious goal of extracting the shared functionality of isoBuffer and isoBuffer_file to a templated container. If I do it right then that container could itself then be made to use a templatized version of isoBufferBuffer (maybe renamed to ringBuffer) to manage storage, but with some extra functionality over top.

EspoTek:
Thanks once again.  I didn't even realise I'd made a second class for the case when a file was being read!  Templating seems like a fantastic idea in this case.
Will review after work tomorrow.  :)

Desktop_Interface/isobuffer.cpp
```diff
@@ -2,93 +2,92 @@
...
-	buffer = (short*) calloc(bufferLen*2, sizeof(short));
-	bufferEnd = bufferLen-1;
-	samplesPerSecond = (double) bufferLen/(double)21;
-	samplesPerSecond = samplesPerSecond/375*VALID_DATA_PER_375;
-	sampleRate_bit = samplesPerSecond * 8;
-	virtualParent = caller;
-	channel = channel_value;
+namespace {
+	static char const * fileHeaderFormat =
+		"EspoTek Labrador DAQ V1.0 Output File\n"
```
EspoTek: I've never thought about spreading text across lines like that, and I've never seen it either.  But now that I have, it makes so much sense to do it that way.

Desktop_Interface/isobuffer.h
```diff
@@ -24,80 +25,89 @@ enum class UartParity : uint8_t;
 // Imagine it as a circular buffer, but with access functions specifically
 // designed for isochronous data from an Xmega.
 
-#define CONSOLE_UPDATE_TIMER_PERIOD (ISO_PACKETS_PER_CTX * 4)
+constexpr auto CONSOLE_UPDATE_TIMER_PERIOD = ISO_PACKETS_PER_CTX * 4;
```
EspoTek: Since this is used only in isoBuffer.cpp, it probably belongs in that namespace.
Also, it scares me a bit putting "auto" on an integral type.  Probably much more predicatable if we just used `uint32_t`

Desktop_Interface/isobuffer.h
```diff
@@ -24,80 +25,89 @@ enum class UartParity : uint8_t;
...
+	QPlainTextEdit* m_console2;
+	unsigned char m_channel = 255;
+	bool m_serialAutoScroll = true;
 
 // Conversion And Sampling
-	double voltage_ref = 1.65;
-	double frontendGain = (R4 / (R3 + R4));
-	int samplesPerSecond;
-	int sampleRate_bit;
+	double m_voltage_ref = 1.65;
```
EspoTek: These 4 variables here are properties of the Labrador board and not this buffer, so I think it makes sense for them to be globals.

EspoTek:
I've squashed it for now and will just manually edit that one file.  Thanks again.  :)

PR #65, opened 2019-01-26
Isobuffer refactor - part 2

SebastianMestre:
Hello there, this one is quite a bit lighter than the previous one. Mostly focused on cleaning up some details that i overlooked or that i deemed would make the previous pr just waay too large.

I further cleaned up readBuffer, I believe it now does what it's suposed to do. It also returns a unique_ptr.

I think isoBuffer can be modified to allocate a buffer half as big as it does now (contiguous range queries are not needed, so a double ring buffer is not needed), but i haven't looked into that just yet since I would also have to study how other classes access m_buffer.

I changed the connect on serialManage to use member function pointers instead of the SIGNAL and SLOT macros. As far as know, this is a best practice and can be done safely and incrementaly all throughout the codebase.

I moved the storage to unique_ptr's. I don't believe there were memory leaks before, even though the buffers weren't being freed on destruction. This is because (it seems to me) all isoBuffers were alive for almost the entire runtime of the application, so the buffers were freed soon after the isoBuffers died anyways. But better safe than sorry. It also led to removing a few calls to free(), which is nice.

There are also quite a few small changes, like changing `m_stopDecoding` to `m_isDecoding` (the motivation was using a positive term for bools, to avoid confusion), reordering member variables (I wanted to have related data close together, mostly for ease of reading), using `==` instead of `>=` on the output file column comparison (to indicate that the column count goes up by one), and a bunch more, related to comments and the like.

### DIFF
--- a/Desktop_Interface/isobuffer.cpp
+++ b/Desktop_Interface/isobuffer.cpp
@@ -25,7 +25,7 @@ namespace
 isoBuffer::isoBuffer(QWidget* parent, int bufferLen, isoDriver* caller, unsigned char channel_value)
 	: QWidget(parent)
 	, m_channel(channel_value)
-	, m_buffer((short*)calloc(bufferLen*2, sizeof(short)))
+	, m_buffer(std::make_unique<short[]>(bufferLen*2))
 	, m_bufferEnd(bufferLen-1)
 	, m_samplesPerSecond(bufferLen/21.0/375*VALID_DATA_PER_375)
 	, m_sampleRate_bit(bufferLen/21.0/375*VALID_DATA_PER_375*8)
@@ -35,7 +35,6 @@ isoBuffer::isoBuffer(QWidget* parent, int bufferLen, isoDriver* caller, unsigned
 
 isoBuffer::~isoBuffer()
 {
-	free(m_buffer);
 	free(m_readData);
 }
 
--- a/Desktop_Interface/isobuffer.cpp
+++ b/Desktop_Interface/isobuffer.cpp
@@ -37,7 +37,7 @@ isoBuffer::isoBuffer(QWidget* parent, int bufferLen, isoDriver* caller, unsigned
 void isoBuffer::insertIntoBuffer(short item)
 {
 	m_buffer[m_back] = item;
-	m_buffer[m_back+m_bufferEnd+1];
+	m_buffer[m_back+m_bufferEnd+1] = item;
 	m_back++;
 	m_insertedCount++;
 
--- a/Desktop_Interface/isobuffer.cpp
+++ b/Desktop_Interface/isobuffer.cpp
@@ -331,10 +331,10 @@ void isoBuffer::serialManage(double baudRate, UartParity parity)
 		connect(m_decoder, &uartStyleDecoder::wireDisconnected,
 		        m_virtualParent, &isoDriver::serialNeedsDisabling);
 	}
-	if (m_stopDecoding)
+	if (!m_isDecoding)
 	{
 		m_decoder->updateTimer->start(CONSOLE_UPDATE_TIMER_PERIOD);
-		m_stopDecoding = false;
+		m_isDecoding = true;
 	}
 	m_decoder->setParityMode(parity);
 	m_decoder->serialDecode(baudRate);
--- a/Desktop_Interface/isobuffer.cpp
+++ b/Desktop_Interface/isobuffer.cpp
@@ -278,6 +278,8 @@ short isoBuffer::inverseSampleConvert(double voltageLevel, int TOP, bool AC) con
 	return sample;
 }
 
+// For capacitance measurement. x0, x1 and x2 are all various time points
+// used to find the RC coefficient.
 template<typename Function>
 int isoBuffer::capSample(int offset, int target, double seconds, double value, Function comp)
 {
@@ -303,8 +305,6 @@ int isoBuffer::capSample(int offset, int target, double seconds, double value, F
 	return -1;
 }
 
-// For capacitance measurement. x0, x1 and x2 are all various time points
-// used to find the RC coefficient.
 int isoBuffer::cap_x0fromLast(double seconds, double vbot)
 {
 	return capSample(0, kSamplesSeekingCap, seconds, vbot, fX0Comp);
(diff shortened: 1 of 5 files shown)


Desktop_Interface/isobuffer.cpp
```diff
@@ -51,49 +57,58 @@ void isoBuffer::insertIntoBuffer(short item)
 
 short isoBuffer::bufferAt(int idx) const
 {
-	return m_buffer[m_back - idx];
+	// NOTE: this is only correct if idx < m_insertedCount
+	return m_buffer[m_back + (m_bufferEnd+1) - idx];
```
EspoTek: Whoa, good catch.  That would have caused hell for high values of idx otherwise.

Desktop_Interface/isobuffer.cpp
```diff
@@ -109,15 +124,13 @@ void isoBuffer::writeBuffer(T* data, int len, int TOP, Function transform)
 	{
 		bool isUsingAC = m_channel == 1
 		                 ? m_virtualParent->AC_CH1
-						 : m_virtualParent->AC_CH2;
+		                 : m_virtualParent->AC_CH2;
 
-		for (int i = 0; i < len; i++)
+		for (int i = 0; i < len && m_fileIOEnabled; i++)
```
EspoTek: Hadn't considered the case where the user aborts the file mid-write.  Nice.

Desktop_Interface/isobuffer.cpp
```diff
@@ -132,57 +145,41 @@ void isoBuffer::writeBuffer_short(short* data, int len)
...
-	 *    - PalmOS
-	 *    - 3BSD
-	 *    - UNIX 7
-	 *   I do not know of any non-compliant somewhat modern implementations.
+	/*
+	 * The expected behavior is to run backwards over the buffer with a stride
+	 * of timeBetweenSamples steps, and push the touched elements into readData.
+	 * If more elements are requested than how many are stored (1), the buffer
+	 * will be populated only partially. Modifying this function to return null
+	 * or a zero-filled buffer insted should be simple enough.
```
EspoTek: insted -> instead

Desktop_Interface/isobuffer.cpp
```diff
@@ -132,57 +145,41 @@ void isoBuffer::writeBuffer_short(short* data, int len)
...
 	 *
-	 * The expected behavior is to cycle backwards over the buffer, taking into
-	 * acount only the part of the buffer that has things stored, with a stride
-	 * of timeBetweenSamples steps, and insert the touched elements into readData.
+	 * (1) m_insertedCount < (delayOffset + sampleWindow) * m_samplesPerSecond
 	 *
-	 * ~Sebastian Mestre
+	 * If this is not the intent, please let me know.
```
EspoTek: That's the intent.  I hadn't considered the case where someone using this function requests an invalid buffer, but your suggestion makes sense.

Desktop_Interface/isobuffer.cpp
SebastianMestre: Ahhh, the typos.  Maybe i should start running my changes through a spell checker. hahaha

Desktop_Interface/isobuffer.cpp
```diff
@@ -132,57 +145,41 @@ void isoBuffer::writeBuffer_short(short* data, int len)
...
 	 */
 	const double timeBetweenSamples = sampleWindow * m_samplesPerSecond / numSamples;
 	const int delaySamples = delayOffset * m_samplesPerSecond;
 
-	free(m_readData);
-
-	m_readData = (short*) calloc(numSamples, sizeof(short));
+	auto readData = std::make_unique<short[]>(numSamples);
```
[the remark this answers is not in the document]
SebastianMestre: Sure thing!

Desktop_Interface/isobuffer.cpp
```diff
@@ -282,6 +278,8 @@ short isoBuffer::inverseSampleConvert(double voltageLevel, int TOP, bool AC) con
 	return sample;
 }
 
+// For capacitance measurement. x0, x1 and x2 are all various time points
+// used to find the RC coefficient.
```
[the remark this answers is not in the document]
SebastianMestre: Haha, not really, that comment was in the file originally, I just moved it to a more prominent spot. I did ask her what an RC coefficient is though!

Desktop_Interface/isodriver.cpp
```diff
@@ -1337,7 +1337,7 @@ double isoDriver::meanVoltageLast(double seconds, unsigned char channel, int TOP
         break;
     }
 
-    short * tempBuffer = currentBuffer->readBuffer(seconds,1024, 0, 0);
+	std::unique_ptr<short[]> tempBuffer = currentBuffer->readBuffer(seconds,1024, 0, 0);
```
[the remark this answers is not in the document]
SebastianMestre: Ah, yes. My next pr could be just running stuff through a code formatter, if you find that useful. (I find that type of pr's annoying since they pollute blame and the commit history, but having consistent formatting can really help readability)

Desktop_Interface/isobuffer.cpp
```diff
@@ -328,15 +325,14 @@ void isoBuffer::serialManage(double baudRate, UartParity parity)
 	if (m_decoder == NULL)
 	{
 		m_decoder = new uartStyleDecoder(this);
-		// TODO: Look into using the type safe version of connect.
```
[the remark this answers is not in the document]
SebastianMestre: Right? I found out about that one the other day watching talk about Qt on youtube, turns out it can catch a bunch of bugs on compile time that would otherwise be crashes (It doesn't seem to be the case here, but better be safe than sorry).

Desktop_Interface/isobuffer.cpp
SebastianMestre: It also stops the loop once the file size limit is reached.

Desktop_Interface/isobuffer.cpp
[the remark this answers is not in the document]
SebastianMestre: I do find that auto works well 99% of the time (exception being how it interacts with std::initializer_list on some corner cases) though.

Desktop_Interface/isobuffer.cpp
SebastianMestre: That's good! Now I can remove that line from the comment, yay!

EspoTek:
Looks good, thanks.

PR #68, opened 2019-03-01
Functiongencontrol refactor

SebastianMestre:
Hello there! It's been a while. I've been kinda busy.

I tried to keep this one at a reasonable size.

I'm not very happy with what I did with functionGenControl::waveformName. (particularly, i feel like the int template parameter could be removed, and the parameters simplified) But I wanted your input on it before I looked into it any further.

I put the various data members into a struct, which i also used in a function in genericusbdriver.cpp. Maybe it should not be inside of the functionGenControl class.

I saw some opportunities to switch over to unique_ptr, but felt it would clutter the pr too much. I think i found a leak as well, I believe i left a comment about it on the relevant spot.

Refactoring this one was a lot easier than isoBuffer since there isn't much tricky index manipulation and the like, and that which there is I didn't touch (apart from renaming things).

Looking forward to hear your feedback. If there is anything in particular you want me to do with this during refactoring, I'd be pleased to do it.


Desktop_Interface/functiongencontrol.cpp
```diff
@@ -5,126 +5,56 @@
 functionGenControl::functionGenControl(QWidget *parent) : QLabel(parent)
 {
     this->hide();
-    samples_CH1 = (unsigned char *) malloc(0);
```
EspoTek: I can only imagine why 2016 me thought it was a good to malloc a 0 byte array?

Desktop_Interface/genericusbdriver.cpp
```diff
@@ -81,54 +81,53 @@ void genericUsbDriver::setFunctionGen(int channel, functionGenControl *fGenContr
        //// CH1 is AUX!!  CH2 is "MAIN"!!////
       //////////////////////////////////////
 
-    int length, maxLength, numDivides, maxDivides;
-    double freq, amplitude, offset;
-    unsigned char *samples;
+    int maxLength; // NOTE: unused
```
EspoTek: Good catch (again, I'm not sure 2016 me thought it was a good idea to avoid compiler warnings!).
This can be deleted, though.

Desktop_Interface/genericusbdriver.cpp
```diff
@@ -81,54 +81,53 @@ void genericUsbDriver::setFunctionGen(int channel, functionGenControl *fGenContr
...
-        offset = offset / 3;
+    if ((channelData.amplitude+channelData.offset) > FGEN_LIMIT){
+        channelData.amplitude = channelData.amplitude / 3;
+        channelData.offset = channelData.offset / 3;
         fGenTriple |= ((unsigned char) !channel + 1);
     }
     else fGenTriple &= ((unsigned char) (254 - !channel));
 
-    //qDebug() << "fGenTriple = " << fGenTriple << "fGenControl = " << fGenControl << "length = " << length << "freq = " << freq << "amplitude = " << amplitude << "offset = " << offset << "samples = " << samples;
+    // qDebug() << "fGenTriple = " << fGenTriple
```
EspoTek: No point leaving in commented out print statements.  They should have been deleted years ago.

Desktop_Interface/genericusbdriver.cpp
```diff
@@ -81,54 +81,53 @@ void genericUsbDriver::setFunctionGen(int channel, functionGenControl *fGenContr
...
+		unsigned char *originalSamples = channelData.samples;
+		channelData.samples = (unsigned char *) malloc(channelData.length);
+		memcpy(channelData.samples, originalSamples, (unsigned int) channelData.length);
+	}
 
     //Triple mode
-    if ((amplitude+offset) > FGEN_LIMIT){
-        amplitude = amplitude / 3;
-        offset = offset / 3;
+    if ((channelData.amplitude+channelData.offset) > FGEN_LIMIT){
```
EspoTek: nit: spaces around the +

Desktop_Interface/genericusbdriver.cpp
```diff
@@ -81,54 +81,53 @@ void genericUsbDriver::setFunctionGen(int channel, functionGenControl *fGenContr
...
     //Waveform scaling in V
     double tempDouble;
-    amplitude = (amplitude * 255) / FGEN_LIMIT;
-    offset = (offset * 255) / FGEN_LIMIT;
-    if (offset<FGEN_OFFSET){
-        if (amplitude>5)
-            amplitude -= FGEN_OFFSET;
+    channelData.amplitude = (channelData.amplitude * 255) / FGEN_LIMIT;
+    channelData.offset = (channelData.offset * 255) / FGEN_LIMIT;
+    if (channelData.offset<FGEN_OFFSET){
```
EspoTek: nit: spaces around <

Desktop_Interface/genericusbdriver.cpp
```diff
@@ -81,54 +81,53 @@ void genericUsbDriver::setFunctionGen(int channel, functionGenControl *fGenContr
...
+        if (channelData.amplitude>5)
```
EspoTek: nit: spaces around >

Desktop_Interface/genericusbdriver.cpp
```diff
@@ -141,38 +140,38 @@ void genericUsbDriver::setFunctionGen(int channel, functionGenControl *fGenContr
 #endif
 
     //Applying amplitude and offset to all samples.
-    for (int i=0;i<length;i++){
-        tempDouble = (double) samples[i];
-        tempDouble *= amplitude;
+    for (int i=0;i<channelData.length;i++){
```
EspoTek: nit: :)

Desktop_Interface/genericusbdriver.cpp
```diff
@@ -81,54 +81,53 @@ void genericUsbDriver::setFunctionGen(int channel, functionGenControl *fGenContr
...
+		memcpy(channelData.samples, originalSamples, (unsigned int) channelData.length);
+	}
 
     //Triple mode
-    if ((amplitude+offset) > FGEN_LIMIT){
-        amplitude = amplitude / 3;
-        offset = offset / 3;
+    if ((channelData.amplitude+channelData.offset) > FGEN_LIMIT){
+        channelData.amplitude = channelData.amplitude / 3;
+        channelData.offset = channelData.offset / 3;
```
EspoTek: Looks like we're dividing doubles by ints here.

Desktop_Interface/genericusbdriver.cpp
```diff
@@ -141,38 +140,38 @@ void genericUsbDriver::setFunctionGen(int channel, functionGenControl *fGenContr
...
-    unsigned char *tempSamples = (unsigned char *) malloc(0);
+    unsigned char *tempSamples = NULL;
 
-    while(length * freq > DAC_SPS){
+    while(channelData.length * channelData.freq > DAC_SPS){
         loop_entered = true;
-        numDivides--;
-        if (numDivides==0){
+        channelData.divisibility--;
+        if (channelData.divisibility==0){
```
EspoTek: nit: spaces around ==

Desktop_Interface/genericusbdriver.cpp
```diff
@@ -141,38 +140,38 @@ void genericUsbDriver::setFunctionGen(int channel, functionGenControl *fGenContr
...
-        length = length >> 1;
+        int shiftTemp = (maxDivides - channelData.divisibility);
+        channelData.length = channelData.length >> 1;
 
         free(tempSamples);
-        tempSamples = (unsigned char *) malloc(length);
-        for (int i=0; i<length;i++){
-            tempSamples[i] = samples[i<<shiftTemp];
+        tempSamples = (unsigned char *) malloc(channelData.length);
+        for (int i=0; i<channelData.length;i++){
```
EspoTek: nit: space around < and { on new line

Desktop_Interface/genericusbdriver.cpp
```diff
@@ -141,38 +140,38 @@ void genericUsbDriver::setFunctionGen(int channel, functionGenControl *fGenContr
...
+            tempSamples[i] = channelData.samples[i<<shiftTemp];
```
EspoTek: nit: space around <<

Desktop_Interface/functiongencontrol.h
```diff
@@ -15,10 +15,27 @@ class functionGenControl : public QLabel
 {
     Q_OBJECT
 public:
+
+	struct ChannelData
+	{
+		unsigned char* samples = NULL;
```
EspoTek: A lot of leaks and memcpy hack in the USB driver could just magically disappear if this was a std::vector.

Performance is not too critical in this instance as the length of samples is under 512 bytes (that's how much RAM I could spare on the microcontroller for function generator data; the two 512-byte buffers take up 1/4 of the XMEGA's 4kB SRAM!), and each change is followed by a blocking call to a libusb API anyway, which transfers these bytes over a 12Mb/s USB link (memory copies on a fairly standard CPU have bandwidth in the single-digit GB/s).

Desktop_Interface/functiongencontrol.cpp
SebastianMestre: Other functions called free() on samples_CH1 and samples_CH2 first thing, it seems this was in place to have something to free on the first call. But since free(NULL) is a no-op, setting it to NULL is perfectly valid, which is what i changed it to.

Desktop_Interface/functiongencontrol.cpp
```diff
@@ -5,126 +5,56 @@
...
+    functionGenToUpdate(1, this);
+}
+
+void functionGenControl::offsetUpdate_CH2(double newOffset){
+    qDebug() << "newOffset2 = " << newOffset;
+	CH2.offset = newOffset;
+    functionGenToUpdate(1, this);
+}
+
+template<int LengthStringSize, typename Lambda>
```
EspoTek: LengthString and DivisibilityString are just small buffers that hold strings to be read in from a file.  The strings should be 4 characters long or less, hence the size of 5.  That being said, you can easily make the length higher (if you look at the Android stuff, it uses 16-byte buffers) and it won't cause any issues.  After all, stack size is something like 8MiB, what difference does 11 bytes make?

Desktop_Interface/functiongencontrol.cpp
```diff
@@ -142,36 +72,37 @@ void functionGenControl::waveformName_CH2(QString newName)
...
+        channel.samples[i] = (unsigned char) dummy;
     }
+
 #else
     QString directory = QCoreApplication::applicationDirPath();
 
     directory.append("/waveforms/");
     directory.append(newName);
-    QByteArray temp = directory.toLatin1();
+    QByteArray temp = encode(directory);
```
EspoTek: This is another parameter than can be removed from the template (soon to be method) - either UTF-8 or the native encoding (which I suspect for most locales would be UTF-8) should always be used as it solves a whole lot of issues related to users having Chinese/Japanese/Korean etc. characters in file names and paths.  At work, it's one of QA's favourite things to do - can we break the software with a Japanese file name?

Desktop_Interface/functiongencontrol.h
SebastianMestre: That's a great point! I don't know how it didn't even cross my mind to use a vector instead.

I got a part time job doing php development recently maybe that's clouding my judgement hahaha

Desktop_Interface/functiongencontrol.cpp
SebastianMestre: I was thinking just that. That change is in!

Desktop_Interface/functiongencontrol.cpp
SebastianMestre: Ahhh that's a good point. I was refactoring without putting much thought into the reason of things so i just parameterized the different functionality and didn't even consider that, but now that you mention it, it feels obvious.

SebastianMestre:
well, to be honest, I didn't do much with that code besides DRYing it up, and cleaning up the formatting on some places. It was pretty clean imo. I guess the two copy-pasted versions could be called sloppy but the code itself seems fine to me.

Yup. std::vector is definitely the way to go. I'm kinda embarrased I didn't think of it. I moved everything over to std::vector on f3e9cab, I'm really pleased with how it improved things.

Wow! I had no idea uinque_ptr introduced such overhead! Surely something I'll watch out for in the future.

Thanks a lot for the feedback! I learned some valuable things!

SebastianMestre:
I'm an idiot. I didnt notice that the functions called at the end of waveformName were different. It's all fixed now. I just threw in an if statement. I feel like the channel could be parameterized through the signal interface (pass in an int or an enum value) instead of using different functions. I'm not completely confident about implementing that myself since I have very limited experience with Qt (also, I ended up expanding this pr way more than i originally intended... again) but I can try if you think that's a good idea.

I'm sorry I keep adding changes after review. I really can't help myself.

Desktop_Interface/functiongencontrol.cpp
```diff
@@ -86,65 +83,103 @@ void functionGenControl::waveformName(QString newName, ChannelData& channel, int
...
 
     int dummy;
     char *dataStringCurrent = dataString;
-    for (int i=0;i<channel.length;i++){
+    for (int i = 0; i < channel.length; i++)
+	{
         sscanf(dataStringCurrent, "%d", &dummy);
         dataStringCurrent += strcspn(dataStringCurrent, "\t") + 1;
-        channel.samples[i] = (unsigned char) dummy;
+        channel.samples[i] = uint8_t(dummy);
```
EspoTek: nit: I usually put the type in brackets (for a C cast), or use `static_cast`

Desktop_Interface/functiongencontrol.cpp
```diff
@@ -86,65 +83,103 @@ void functionGenControl::waveformName(QString newName, ChannelData& channel, int
     QByteArray remainingData = fptr.readAll();
     char *dataString = remainingData.data();
 
-    free(channel.samples);
-    channel.samples = (unsigned char *) malloc(channel.length);
+	channel.samples.resize(channel.length);
```
EspoTek: I don't think it makes sense to store the length as a member variable in the channels struct, since we now have a `std::vector` with a `size()`method

Desktop_Interface/functiongencontrol.cpp
```diff
@@ -86,65 +83,103 @@ void functionGenControl::waveformName(QString newName, ChannelData& channel, int
...
 
 #else
-    QString directory = QCoreApplication::applicationDirPath();
 
-    directory.append("/waveforms/");
-    directory.append(newName);
-    QByteArray temp = encode(directory);
-    char *fileName = temp.data();
+    QByteArray filePath = QCoreApplication::applicationDirPath()
+		.append("/waveforms/").append(newName).toUtf8();
```
EspoTek: I would be wary about passing a UTF-8 array to `fopen()`.  It may work, but I wouldn't trust it to be cross-platform and safe in all locales.  Local 8-bit should be much safer.

Desktop_Interface/functiongencontrol.cpp
```diff
@@ -86,65 +83,103 @@ void functionGenControl::waveformName(QString newName, ChannelData& channel, int
...
-    setMinFreq_CH1((double) CLOCK_FREQ/1024/65535/channel.length);
-    functionGenToUpdate(channelID, this); // 1ST ARG = 0
+	double newMaxFreq = DAC_SPS / (channel.length >> (channel.divisibility - 1));
+	double newMinFreq = double(CLOCK_FREQ) / 1024.0 / 65535.0 / channel.length;
+
+	// NOTE: Not very clean... Not sure what to do about it.
+	// I guess the "right thing" would be to have a Channel QObject class with its
+	// own signals and slots, or have a single setMaxFreq signal with channelID as
+	// an argument. Either solution would require changes in other places in the
+	// codebase so this will have to do for now.
```
EspoTek: Haha, that honestly that isn't the worst thing in the world.  Yes, it's an unnecessary if-else block but there's so much worse on this Earth.

Desktop_Interface/functiongencontrol.h
```diff
@@ -18,7 +20,8 @@ class functionGenControl : public QLabel
 
 	struct ChannelData
 	{
-		unsigned char* samples = NULL;
+		std::vector<uint8_t> samples;
+		// TODO: get rid of length member. samples:std::vector already has this information
```
EspoTek: Ahh, just read the TODO.  Ignore comment above!  :)

Desktop_Interface/functiongencontrol.h
```diff
@@ -28,13 +31,15 @@ class functionGenControl : public QLabel
...
 
-	ChannelData CH1;
-	ChannelData CH2;
+	ChannelData channels[2];
 
 private:
 
-template<int LengthStringSize, typename Lambda>
-void waveformName(QString newName, ChannelData& channel, int channelID, Lambda encode);
+	// NOTE: An enum instead of a plain int would probably be better here
```
EspoTek: Hadn't thought of that but that's a fantastic idea!

Desktop_Interface/genericusbdriver.cpp
```diff
@@ -139,67 +127,63 @@ void genericUsbDriver::setFunctionGen(int channel, functionGenControl *fGenContr
...
 
-    //Applying amplitude and offset to all samples.
-    for (int i=0;i<channelData.length;i++){
-        tempDouble = (double) channelData.samples[i];
-        tempDouble *= channelData.amplitude;
-        tempDouble /= 255;
-        tempDouble += channelData.offset;
-        channelData.samples[i] = (unsigned char) tempDouble;
-    }
+	auto applyAmplitudeAndOffset = [=](unsigned char sample) -> unsigned char
```
EspoTek: Wow, that's a bit neater!

EspoTek:
Hey, congrats on the job.  I wish I could have found one while I was still studying, honestly.  Even if you're not doing the exact thing you plan on doing when you graduate, just learning how to work in that sort of business environment is an immensely useful skill to have, and having something on your resume makes you 1000x more attractive than having nothing.

I was talking to a friend recently who was involved in the hiring process for a server backend dev job and they basically just brushed over all of the applications with no experience/portfolio.  Which is understandable, too - if someone was capable of programming and has access to a computer they probably should have done _something_ while studying.  Or you have to be really good at bullshitting.  :)

I've merged in your PR for the moment since the only big comments I had to make were already in as TODOs.  Thanks once again for everything.

PR #69, opened 2019-03-05
Functiongencontrol refactor - part 2

SebastianMestre:
Hello there! this is a follow-up on the previous pr. Just tidying up some details.

Not much going on, really.

### DIFF
--- a/Desktop_Interface/functiongencontrol.cpp
+++ b/Desktop_Interface/functiongencontrol.cpp
@@ -9,49 +9,54 @@ functionGenControl::functionGenControl(QWidget *parent) : QLabel(parent)
 
 void functionGenControl::waveformName_CH1(QString newName)
 {
-	waveformName(newName, 0);
+	waveformName(ChannelID::CH1, newName);
 }
 
 void functionGenControl::freqUpdate_CH1(double newFreq)
 {
-	freqUpdate(newFreq, 0);
+	freqUpdate(ChannelID::CH1, newFreq);
 }
 
 void functionGenControl::amplitudeUpdate_CH1(double newAmplitude)
 {
-	amplitudeUpdate(newAmplitude, 0);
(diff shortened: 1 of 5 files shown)


Desktop_Interface/functiongencontrol.cpp
```diff
@@ -119,67 +126,68 @@ void functionGenControl::waveformName(QString newName, int channelID)
...
     }
 
     free(dataString);
     fclose(fptr);
 #endif
 
-	double newMaxFreq = DAC_SPS / (channel.length >> (channel.divisibility - 1));
-	double newMinFreq = double(CLOCK_FREQ) / 1024.0 / 65535.0 / channel.length;
+	double newMaxFreq = DAC_SPS / (length >> (channel.divisibility - 1));
+	double newMinFreq = double(CLOCK_FREQ) / 1024.0 / 65535.0 / length;
```
EspoTek: Hey man.  Once again, this is my original bad, but it looks like length is not cast to double here.  Float / int = int.

Desktop_Interface/functiongencontrol.h
```diff
@@ -17,37 +17,42 @@ class functionGenControl : public QLabel
...
 		std::vector<uint8_t> samples;
-		// TODO: get rid of length member. samples:std::vector already has this information
-		int length;
 		int divisibility;
 		double freq = 1000.0;
 		double amplitude = 0.0;
 		double offset = 0.0;
 	};
 
+	enum class ChannelID
```
EspoTek: This is a really good thing that I think makes sense to use in many, many more places in the code.  For now, leave it here, but I think in future putting it in one of the global includes would be a good idea.

EspoTek:
Good job once again.
There was one pretty minor thing I noticed (which I introduced!), but apart from that it all makes sense.  I've merged it and will just add an extra tiny commit later.

PR #71, opened 2019-03-07
Access the pointer directly on isoBufferBuffer::insert

SebastianMestre:
It came to mind after what you told me about the other place where i used unique_ptr.

I ran some benchmarks on this and `isoBufferBuffer::insert(const char*)` ended up being 2x faster than accessing the unique_ptr directly, and about 20% slower than a raw pointer with optimizer set to -O0 or -O1.

On higher optimization levels, the difference between the 3 different methods is too small to measure. This is likely due to the compiler moving the unique_ptr::get out of the loop, removing the overhead almost completely.

PS: I have a question about a particular piece of code somewhere else on the codebase. Is there a "right place" for me to ask about it? I figured spamming your email with questions without asking if that's ok first would not be a nice thing to do.

### DIFF
--- a/Desktop_Interface/isobufferbuffer.cpp
+++ b/Desktop_Interface/isobufferbuffer.cpp
@@ -31,10 +31,12 @@ isoBufferBuffer::isoBufferBuffer(uint32_t length)
 // Adds a character to the end of the buffer
 void isoBufferBuffer::insert(char c)
 {
+	char* dataPtr = m_data.get();
+
 	// Add character to first half of the buffer
-	m_data[m_top] = c;
+	dataPtr[m_top] = c;
 	// Then to the second
-	m_data[m_top+m_capacity] = c;
+	dataPtr[m_top+m_capacity] = c;
 
 	// Loop the buffer index if necessary and update size accordingly
 	m_top = (m_top + 1) % m_capacity;


EspoTek:
Hmm, that's interesting, actually.  Perhaps I was accidentally running the code with -O0 when doing my test?  It's certainly possible.  In the .pro file I mess with the flags a bit (and I'm not proud of this looking back, but one of those flags is to force O0 on ARM Linux as that magically fixes one of the bugs), or perhaps I was running a debug build without noticing?  I do remember writing the patch after a whole day of programming and a glass of wine, so perhaps I just didn't realise?

Email is probably the best place to ask questions about the code.  And definitely wouldn't call a question or two about the code "spamming".  I mean, sure, if you send a dozen emails in one day that might be a bit much, but general questions about the code is honestly appreciated.

You look a lot less intimidating in the new picture, by the way.  ;)

Cheers,
~Chris

SebastianMestre:
> Perhaps I was accidentally running the code with -O0

That might be it (i just re-ran my benchmarks, just to be sure; it also turns out the cost of indexing into unique_ptr in -O1, which i had not measured, is very small: around 30% slower than this pr on -O1). Or maybe it's because we're running different compilers? Anyways, after changing this around it seems like the heaviest operation is the modulo after each insert(char), and the fact that the following operations depend on that being done, so there can be no Instruction Level Parallelism.

Taking that into account, i wrote a version of this function that is 30x faster (in my one benchmark that doesn't really represent the real use case). It's kinda ugly and unnecesary (it seems to me like isoBufferBuffer is mostly used to store logs and its performance isn't critical) but it was a fun excercise.

Here's a pastebin if you're interested: https://pastebin.com/eqKy9nG4

 > Email is probably the best place to ask questions about the code.

Ok! I'll send you an email later today or tomorrow.

 > You look a lot less intimidating in the new picture, by the way. ;)

Hahahaha. Thank you! I find that I tend to look silly when I smile for a picture, so I mostly use pictures where I have a straight face.

PR #72, opened 2019-03-10
Uartstyledecoder refactor

SebastianMestre:
Hello!

Nothing too crazy on this one. Just a bunch of code transformations that hopefully don't break anything. Biggest thing is definitely making `serialBuffer:isoBufferBuffer` and `updateTimer:QTimer` part of the class instead of pointers (which allowed me to get rid of the body of the destructor). I tried comparing the behavior of some of the methods before and after with some test cases and it seems the same. I might be wrong though, it's 3am and I am having a hard time wrapping my head around the state space of this class.

There will definitely two or three more prs refactoring this class: I'm trying to dial back the size of my prs, which have a tendency to get huge, so I figured I would spread my work over a greater amount of smaller prs. This means some things were left "unfinished" (not really, everything still works but the refactor is less than half done)

The general trend in this pr is making methods act more like functions instead of procedures (i.e. take arguments and return a value instead of using shared state to exchange information). This is something that is still unfinished, a few more of the methods can be made to act more like functions, I just haven't gotten around to it yet.

The style of the code in this file is very different than what i saw on, for instance, isoBuffer. way cleaner, way less duplication, a lot easier to refactor. I wonder if it was written after isoBuffer?

This week is going to be a busy one for me so expect the next pr somewhere around next weekend.

### DIFF
--- a/Desktop_Interface/uartstyledecoder.cpp
+++ b/Desktop_Interface/uartstyledecoder.cpp
@@ -233,16 +233,12 @@ void uartStyleDecoder::performParityCheck()
 {
     auto isEvenParity = [=] () -> bool
     {
-        uint32_t mask = 0x00000001;
-        uint8_t parity = 0;
-        for (int i = 0; i < dataBit_max; i++)
-        {
-            const uint8_t currentBit = (dataBit_current & mask) ? 1 : 0;
-            parity = parity ^ currentBit;
-            mask = mask << 1;
-        }
+		bool result = false;
 
-        return parity == 0;
+		for (uint32_t mask = 1 << (dataBit_max-1); mask != 0; mask >>= 1)
+			result ^= static_cast<bool>(dataBit_current & mask);
+
+		return not result;
     };
 
     switch(parity)
--- a/Desktop_Interface/uartstyledecoder.cpp
+++ b/Desktop_Interface/uartstyledecoder.cpp
@@ -132,7 +132,7 @@ void uartStyleDecoder::decodeNextUartBit(unsigned char bitValue)
     if (dataBit_current == parityIndex)
     {
         assert(parity != UartParity::None);
-        performParityCheck();
+        parityCheckFailed = not isParityCorrect(dataBit_current);
         dataBit_current++;
     }
     else if (dataBit_current < dataBit_max)
@@ -230,31 +230,21 @@ void uartStyleDecoder::setParityMode(UartParity newParity)
     parity = newParity;
 }
 
-void uartStyleDecoder::performParityCheck()
+bool uartStyleDecoder::isParityCorrect(uint32_t bitField) const
 {
-    auto isEvenParity = [=] (uint32_t bitField) -> bool
-    {
-		bool result = false;
-
-		for (uint32_t mask = 1 << (dataBit_max-1); mask != 0; mask >>= 1)
-			result ^= static_cast<bool>(bitField & mask);
-
-		return not result;
-    };
-
-    switch(parity)
-    {
-    case UartParity::None:
+	if (parity == UartParity::None)
 		std::terminate();
-        break;
-    case UartParity::Even:
-        parityCheckFailed = not isEvenParity(dataBit_current);
-		break; // NOTE: There used to be a fallthrough here but it seemed like a bug
-    case UartParity::Odd:
-        parityCheckFailed = isEvenParity(dataBit_current);
-		break;
-    }
 
-    return;
+	return parityOf(bitField) == parity;
+}
+
+UartParity uartStyleDecoder::parityOf(uint32_t bitField) const
+{
+	bool result = false;
+
+	for (uint32_t mask = 1 << (dataBit_max-1); mask != 0; mask >>= 1)
+		result ^= static_cast<bool>(bitField & mask);
+
+	return result ? UartParity::Odd : UartParity::Even;
 }
 
--- a/Desktop_Interface/uartstyledecoder.cpp
+++ b/Desktop_Interface/uartstyledecoder.cpp
@@ -49,9 +49,6 @@ void uartStyleDecoder::updateConsole()
 
 void uartStyleDecoder::serialDecode(double baudRate)
 {
-    /*if(stopDecoding){
-        return;
-    }*/
     double dist_seconds = (double)serialDistance()/(parent->m_sampleRate_bit);
     double bitPeriod_seconds = 1/baudRate;
 
@@ -61,7 +58,7 @@ void uartStyleDecoder::serialDecode(double baudRate)
     while(dist_seconds > (bitPeriod_seconds + SERIAL_DELAY))
 	{
         // Read next uart bit
-        unsigned char uart_bit = getNextUartBit();
+        unsigned char uart_bit = getUartBit(serialPtr_bit);
 
         if (uart_bit == 1)
             allZeroes = false;
@@ -118,10 +115,10 @@ void uartStyleDecoder::updateSerialPtr(double baudRate, unsigned char current_bi
         serialPtr_bit -= (parent->m_bufferLen * 8);
 }
 
-unsigned char uartStyleDecoder::getNextUartBit() const
+unsigned char uartStyleDecoder::getUartBit(int bitIndex) const
 {
-    int coord_byte = serialPtr_bit/8;
-    int coord_bit = serialPtr_bit - (8*coord_byte);
+    int coord_byte = bitIndex/8;
+    int coord_bit = bitIndex - (8*coord_byte);
     unsigned char dataByte = parent->m_buffer[coord_byte];
     unsigned char mask = (0x01 << coord_bit);
     return ((dataByte & mask) ? 1 : 0);
@@ -142,7 +139,16 @@ void uartStyleDecoder::decodeNextUartBit(unsigned char bitValue)
     }
     else
     {
-        decodeDatabit(dataBit_max + 1);
+        char decodedDatabit = decodeDatabit(dataBit_max + 1, currentUartSymbol);
+
+		if (parityCheckFailed)
+		{
+			m_serialBuffer.insert("\n<ERROR: Following character contains parity error>\n");
+			parityCheckFailed = false;
+		}
+
+		m_serialBuffer.insert(decodedDatabit);
+
         currentUartSymbol = 0;
         dataBit_current = 0;
         uartTransmitting = false;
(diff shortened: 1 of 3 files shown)


Desktop_Interface/uartstyledecoder.cpp
```diff
@@ -2,41 +2,43 @@
 #include <QDebug>
 #include <cassert>
 
-uartStyleDecoder::uartStyleDecoder(QObject *parent_in) : QObject(parent_in)
+uartStyleDecoder::uartStyleDecoder(QObject *parent_in)
+	: QObject(parent_in)
+	, parent{static_cast<isoBuffer*>(parent_in)}
```
EspoTek: If parent is always an `isoBuffer`, could we just use the `isoBuffer()` constructor and get rid of the `QObject` stuff completely?

Desktop_Interface/uartstyledecoder.cpp
```diff
@@ -2,41 +2,43 @@
...
+    m_updateTimer.setTimerType(Qt::PreciseTimer);
+    m_updateTimer.start(CONSOLE_UPDATE_TIMER_PERIOD);
+    connect(&m_updateTimer, &QTimer::timeout, this, &uartStyleDecoder::updateConsole);
 
-    serialBuffer = new isoBufferBuffer(SERIAL_BUFFER_LENGTH);
-
-    if(parent->m_channel == 1) console = parent->m_console1;
-    else if(parent->m_channel == 2) console = parent->m_console2;
-    else qFatal("Nonexistant console requested in uartStyleDecoder::serialDecode");
+    if (parent->m_channel == 1)
```
EspoTek: This would need to be a whole other PR, but rather than constantly reading the internal state of the parent, many of these variables should be passed in as constructor arguments.  The one or two that need to be polled should be functions in isoBuffer.

Desktop_Interface/uartstyledecoder.cpp
```diff
@@ -2,41 +2,43 @@
...
+		console = parent->m_console1;
+    else if (parent->m_channel == 2)
+		console = parent->m_console2;
+    else
+		qFatal("Nonexistant console requested in uartStyleDecoder::serialDecode");
 }
 
 uartStyleDecoder::~uartStyleDecoder()
 {
-    std::lock_guard<std::mutex> lock(mutex);
```
EspoTek: If the destructor is empty now, we don't need to define it at all.  Just `default` it in the header or don't bother at all.

Desktop_Interface/uartstyledecoder.cpp
```diff
@@ -2,41 +2,43 @@
 #include <QDebug>
 #include <cassert>
 
-uartStyleDecoder::uartStyleDecoder(QObject *parent_in) : QObject(parent_in)
+uartStyleDecoder::uartStyleDecoder(QObject *parent_in)
+	: QObject(parent_in)
```
EspoTek: nit: `m_parent` and `parent` would be better names than `parent` and `parent_in`.

That said, I don't think we should store parent at all (see below).

Desktop_Interface/uartstyledecoder.cpp
```diff
@@ -2,41 +2,43 @@
...
 }
 
-void uartStyleDecoder::updateConsole(){
+void uartStyleDecoder::updateConsole()
+{
+	if (!newUartSymbol)
+		return;
+
     std::lock_guard<std::mutex> lock(mutex);
-	if(!newUartSymbol) return;
```
EspoTek: Oh man, I cringed so hard reading the original code here.  Who locks a mutex _before_ checking a bool!?

Desktop_Interface/uartstyledecoder.cpp
```diff
@@ -146,29 +157,30 @@ void uartStyleDecoder::decodeNextUartBit(unsigned char bitValue)
...
-    if (left_coord < 0){
+    if (left_coord < 0)
         return true; //Don't want to read out of bounds!!
-    }
 
     //The usual case, when transmitting anyway.
     unsigned char left_byte = (parent->m_buffer[left_coord/8] & 0xff);
     //Only run when a zero is detected in the leftmost symbol.
-    if(left_byte != 255){
+    if (left_byte != 255)
```
EspoTek: nit: as this is doing a binary comparison, it should be testing against 0xff

Desktop_Interface/uartstyledecoder.cpp
```diff
@@ -180,27 +192,24 @@ bool uartStyleDecoder::jitterCompensationProcedure(double baudRate, unsigned cha
...
-    if (parityCheckFailed)
-    {
-        serialBuffer->insert("\n<ERROR: Following character contains parity error>\n");
-        parityCheckFailed = false;
-    }
-    serialBuffer->insert(tempchar);
 }
 
-char uartStyleDecoder::decode_baudot(short symbol){
+char uartStyleDecoder::decodeBaudot([[maybe_unused]] short symbol) const
```
EspoTek: I wouldn't suppress this warning here - there should be a switch/case statement here enumerating all the different values for baudot encoding.  It's just an empty function for now as this feature is unimplemented and so constantly getting a reminder every time I compile is a good thing.  :)

Desktop_Interface/uartstyledecoder.cpp
```diff
@@ -219,35 +228,21 @@ void uartStyleDecoder::setParityMode(UartParity newParity)
...
-    case UartParity::None:
-        assert(false);
-        return;
-    case UartParity::Even:
-        parityCheckFailed = ! isEvenParity();
-    case UartParity::Odd:
-        parityCheckFailed = isEvenParity();
-    }
+	if (parity == UartParity::None)
+		std::terminate();
```
EspoTek: Whoa, a full-on `terminate()` might be a bit heavy-handed here!  Just an `assert(parity != UartParity::None)` will catch the potential error during debugging without blowing up the whole software.

Desktop_Interface/uartstyledecoder.cpp
```diff
@@ -219,35 +228,21 @@ void uartStyleDecoder::setParityMode(UartParity newParity)
...
 
-    return;
+	return parityOf(bitField) == parity;
 }
 
+UartParity uartStyleDecoder::parityOf(uint32_t bitField) const
+{
+	bool result = false;
```
EspoTek: I'm not sure that `result` is the best name for the bool here.  Maybe something like `isOddParity`?

Desktop_Interface/uartstyledecoder.cpp
```diff
@@ -70,52 +70,55 @@ void uartStyleDecoder::serialDecode(double baudRate)
...
-    if (serialPtr_bit >= (parent->m_bufferLen * 8)){
+    if (serialPtr_bit >= (parent->m_bufferLen * 8))
         serialPtr_bit -= (parent->m_bufferLen * 8);
-    }
 }
 
-unsigned char uartStyleDecoder::getNextUartBit(){
-    int coord_byte = serialPtr_bit/8;
-    int coord_bit = serialPtr_bit - (8*coord_byte);
+unsigned char uartStyleDecoder::getUartBit(int bitIndex) const
```
EspoTek: I understand the intention here, but I think this function makes sense more in its original form.  This whole class is basically a monolithic state machine that decodes a (virtual representation of a) UART stream, so random access to the buffer doesn't quite make sense.  In real life, the bits from a UART stream would be buffered in a FIFO and correct symbol decoding depends on getting them in this order.

Desktop_Interface/uartstyledecoder.cpp
```diff
@@ -146,29 +157,30 @@ void uartStyleDecoder::decodeNextUartBit(unsigned char bitValue)
 
 //This function compensates for jitter by, when the current bit is a "1", and the last bit was a zero, setting the pointer
 //to the sample at the midpoint between this bit and the last.
-bool uartStyleDecoder::jitterCompensationProcedure(double baudRate, unsigned char current_bit){
+bool uartStyleDecoder::jitterCompensationProcedure(double baudRate, unsigned char current_bit)
+{
 
     //We only run when the current bit is a "1", to prevent slowdown when there are long breaks between transmissions.
```
EspoTek: Wow, not commenting on your code but this is a giant, giant oversight on my part.  A long stream of binary `1` on a UART indicates idle state between transmissions, not `0`.

This code, as it stands, will iterate over the entire idle period until it finds either the start of the buffer or the last `0` from the previous transmission.
An older version of the code used to grind to a halt when you disconnected the UART wire.  The "fix" was to detect this state and disable the decoder through the `wireDisconnected` signal.

A `while` loop is also not the best way to handle this, as it won't give up in the case of an invalid signal.  I think I'll need to have another look at this method when I find the time.

Desktop_Interface/uartstyledecoder.h
```diff
@@ -34,17 +34,27 @@ class uartStyleDecoder : public QObject
     uint32_t dataBit_max = 7;
     unsigned short currentUartSymbol = 0;
     bool jitterCompensationNeeded = true;
+
     void updateSerialPtr(double baudRate, unsigned char current_bit);
-    unsigned char getNextUartBit();
+    unsigned char getUartBit(int bitIndex) const;
     void decodeNextUartBit(unsigned char bitValue);
```
EspoTek: I'd recommend keeping this as a state machine too.  That's what makes the most sense, given the FIFOness of it.

Desktop_Interface/uartstyledecoder.cpp
SebastianMestre: I feel like unused argument is not the right warning though. Maybe adding a #warning would be better (it's not standard afaik but i think most compilers support it)?

Desktop_Interface/uartstyledecoder.cpp
SebastianMestre: doesn't assert terminate too?

Desktop_Interface/uartstyledecoder.cpp
SebastianMestre: Got it. I just read up on what UART is, I see your point

Desktop_Interface/uartstyledecoder.cpp
SebastianMestre: I'll remove the anotation for now

Desktop_Interface/uartstyledecoder.cpp
SebastianMestre: hahaha I have close to zero experience working with parallelism and synchronization, so I was really scared I would break something when i wrote this diff. I'm really happy I did the right thing!

EspoTek: The lock_guard goes out of scope as soon as the function returns - you had nothing to worry about!
Now, if I'd used the `lock` method of std::mutex this would be a very different story (and you'd have probably fixed a bug).

EspoTek:
Looks much cleaner now.

As mentioned before, I still think it would be best to remove `m_parent`, but that would belong in another PR.  I'll merge this one for now.

PR #73, opened 2019-03-11
Updated Win32 to build with Qt 5.12 on VS 2017

EspoTek:
test


EspoTek:
Submit when bin repo updated to ai157.

PR #74, opened 2019-03-18
'Functiongencontrol refactor - part 3

SebastianMestre:
Once again, I did my best to keep this one manageable.

I moved the `enum class ChannelId` outside of `functionGenControl`, same thing with `struct ChannelData`. I just left them there for now but since they (particularly ChannelID) are used on a bunch of places (and i think using them in more places, like in isoBuffer, would be beneficial), right beside `functionGenControl` might not be the best place to have them.

I also did a few more changes that became obvious once `ChannelData` was its own type, like have `genericUsbDriver` point to a single `ChannelData` instead of pointing to a `functionGenControl` and access the right data depending on the channel.

The functionGenControl code actually became a bit more verbose due to changes in the functionGenToUpdate singnal, but that can be remedied by moving that functionality into `ChannelData`, which is kinda what this refactor was leading up as an unconscious end goal. Maybe next pr? We'll see. Until then.

Best regards, Sebastian.

### DIFF
--- a/Desktop_Interface/functiongencontrol.cpp
+++ b/Desktop_Interface/functiongencontrol.cpp
@@ -167,27 +167,30 @@ void functionGenControl::waveformName(ChannelID channelID, QString newName)
 		break;
 	}
 
-    functionGenToUpdate(channelID, this);
+    functionGenToUpdate(channelID, &channel);
 }
 
 void functionGenControl::freqUpdate(ChannelID channelID, double newFreq)
 {
 	qDebug() << "newFreq" << int(channelID) << " = " << newFreq;
-	getChannelData(channelID).freq = newFreq;
-	functionGenToUpdate(channelID, this);
+	ChannelData& channel = getChannelData(channelID);
+	channel.freq = newFreq;
+	functionGenToUpdate(channelID, &channel);
 }
 
 void functionGenControl::amplitudeUpdate(ChannelID channelID, double newAmplitude)
 {
 	qDebug() << "newAmplitude" << int(channelID) << " = " << newAmplitude;
-	getChannelData(channelID).amplitude = newAmplitude;
-	functionGenToUpdate(channelID, this);
+	ChannelData& channel = getChannelData(channelID);
+	channel.amplitude = newAmplitude;
+	functionGenToUpdate(channelID, &channel);
 }
 
 void functionGenControl::offsetUpdate(ChannelID channelID, double newOffset)
 {
 	qDebug() << "newOffset" << int(channelID) << " = " << newOffset;
-	getChannelData(channelID).offset = newOffset;
-	functionGenToUpdate(channelID, this);
+	ChannelData& channel = getChannelData(channelID);
+	channel.offset = newOffset;
+	functionGenToUpdate(channelID, &channel);
 }
 
--- a/Desktop_Interface/functiongencontrol.cpp
+++ b/Desktop_Interface/functiongencontrol.cpp
@@ -1,63 +1,14 @@
 #include "functiongencontrol.h"
 #include "platformspecific.h"
 
+namespace functionGen {
 
-functionGenControl::functionGenControl(QWidget *parent) : QLabel(parent)
-{
-    this->hide();
+ChannelData const& SingleChannelController::getData() const {
+	return m_data;
 }
 
-void functionGenControl::waveformName_CH1(QString newName)
+void SingleChannelController::waveformName(QString newName)
 {
-	waveformName(ChannelID::CH1, newName);
-}
-
-void functionGenControl::freqUpdate_CH1(double newFreq)
-{
-	freqUpdate(ChannelID::CH1, newFreq);
-}
-
-void functionGenControl::amplitudeUpdate_CH1(double newAmplitude)
-{
-	amplitudeUpdate(ChannelID::CH1, newAmplitude);
-}
-
-void functionGenControl::offsetUpdate_CH1(double newOffset)
-{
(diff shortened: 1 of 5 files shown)


Desktop_Interface/functiongencontrol.cpp
```diff
@@ -167,27 +167,30 @@ void functionGenControl::waveformName(ChannelID channelID, QString newName)
...
-    functionGenToUpdate(channelID, this);
+    functionGenToUpdate(channelID, &channel);
 }
 
 void functionGenControl::freqUpdate(ChannelID channelID, double newFreq)
 {
 	qDebug() << "newFreq" << int(channelID) << " = " << newFreq;
-	getChannelData(channelID).freq = newFreq;
-	functionGenToUpdate(channelID, this);
+	ChannelData& channel = getChannelData(channelID);
```
EspoTek: I'd name this `channelData`.  `channel` implies that it's the ID number.

Desktop_Interface/functiongencontrol.cpp
```diff
@@ -167,27 +167,30 @@ void functionGenControl::waveformName(ChannelID channelID, QString newName)
...
+	channel.freq = newFreq;
+	functionGenToUpdate(channelID, &channel);
 }
 
 void functionGenControl::amplitudeUpdate(ChannelID channelID, double newAmplitude)
 {
 	qDebug() << "newAmplitude" << int(channelID) << " = " << newAmplitude;
-	getChannelData(channelID).amplitude = newAmplitude;
-	functionGenToUpdate(channelID, this);
+	ChannelData& channel = getChannelData(channelID);
```
EspoTek: ditto

Desktop_Interface/functiongencontrol.cpp
```diff
@@ -167,27 +167,30 @@ void functionGenControl::waveformName(ChannelID channelID, QString newName)
...
+	channel.amplitude = newAmplitude;
+	functionGenToUpdate(channelID, &channel);
 }
 
 void functionGenControl::offsetUpdate(ChannelID channelID, double newOffset)
 {
 	qDebug() << "newOffset" << int(channelID) << " = " << newOffset;
-	getChannelData(channelID).offset = newOffset;
-	functionGenToUpdate(channelID, this);
+	ChannelData& channel = getChannelData(channelID);
```
EspoTek: ditto

Desktop_Interface/functiongencontrol.h
```diff
@@ -13,25 +13,26 @@
 
 //functionGenControl is a centralised object to control all of the high-level function gen commands for both channels.
 
+enum class ChannelID
```
EspoTek: I know I said it would be best to move these out of the `functionGenControl`, but looking at it again I think it actually makes sense to keep them  namespaced.  "Channel", in the context of the Labrador, could refer to a signal generator, oscilloscope or logic analyzer channel, and all of these have different data.

Desktop_Interface/genericusbdriver.cpp
```diff
@@ -87,13 +86,15 @@ void genericUsbDriver::setFunctionGen(functionGenControl::ChannelID channelID, f
...
-    if (channelID == ChannelID::CH1)
-		fGenPtr_CH1 = fGenControl;
-    else
-		fGenPtr_CH2 = fGenControl;
+	fGenPtrData[(int)channelID] = fGenControlData;
+
+	useFunctionGenData(channelID);
+}
 
+void genericUsbDriver::useFunctionGenData(ChannelID channelID)
```
EspoTek: I'm not sure "use" is the best word here.  Maybe "send"?

SebastianMestre: Ah yes, i though i had left a comment about that, I wasn't really sure what to name it.

Desktop_Interface/functiongencontrol.h
SebastianMestre: How about adding a `functionGen` namespace and renaming the `functionGenControl` class to `Controller`?

It is not a good fit with how the rest of the codebase is structured but I think it's worth a shot.

Then, we could have `functionGen::Controller`, `functionGen::ChannelID`, and `functionGen::ChannelData` which, inside the namespace would just be `Controller`, `ChannelID`, and `ChannelData`.

This looks like a good pattern to me because there are other places in the codebase that could use a `ChannelID` that shouldn't be implicitly convertible to `functionGen::ChannelID` but are otherwise identical.

There is also the thing that you cant do `using class::member_type;` like you can do with explicit namespaces, which then leads to awkwardness like `using member_type = class::member_type;`

SebastianMestre:
Hello! This last diff is a pretty interesting one, I'm not sure what you'll think of it. I feel like since the change is about organization, it might be more subjective than previous prs that were more clear cut.

I look forward to hearing your opinion.

Desktop_Interface/functiongencontrol.h
```diff
@@ -29,39 +30,66 @@ struct ChannelData
...
     void amplitudeUpdate_CH2(double newAmplitude);
     void offsetUpdate_CH2(double newOffset);
+
+private:
+	SingleChannelController m_channels[2];
 };
 
+}
+
+using functionGenControl = functionGen::DualChannelController;
```
EspoTek: Understand why you did this (makes testing a whole lot easier), but as a permanent solution there's a little bit of a smell there.

SebastianMestre: Of course, I just didn't want this diff to force me to change many different files unnecesarily

PR #76, opened 2019-03-28
I2cdecoder refactor

SebastianMestre:
Code looks pretty darn good around here so there wasn't much to do.

Summary:
 - Constructor new'd up an isoBufferBuffer and a QTimer that were both never freed. I just went for the easy solution and added a destructor. Maybe storing by value or using unique_ptr would be more appropiate?
 - Constructor used SIGNAL and SLOT macros, i replaced that by the type-safe member function pointer overload.
 - Some debug print was left behind, I removed it.

Some things I noticed (but didn't act upon) are:
 - Raw pointers to i2cDecoder::serialBuffer and i2cDecoder::updateTimer (could be store by value or maybe unique_ptr)
 - Everything is public (not really an issue since seemingly no other classes take advantage of this)
 - There is some inconsistent formatting (I didn't want to mess up the blame though so i didn't touch it)

That's all I noticed.

Sebastian.

### DIFF
--- a/Desktop_Interface/i2cdecoder.cpp
+++ b/Desktop_Interface/i2cdecoder.cpp
@@ -86,14 +86,6 @@ void i2cDecoder::runStateMachine()
     edge sdaEdge = edgeDetection(currentSdaValue, previousSdaValue);
 	edge sclEdge = edgeDetection(currentSclValue, previousSclValue);
 
-//    if (sdaEdge == edge::rising || sdaEdge == edge::falling)
-//        qDebug() << "sdaEdge";
-//    if (sclEdge == edge::rising || sclEdge == edge::falling)
-//        qDebug() << "sclEdge";
-
-//    if (sclEdge != edge::held_low)
-//        qDebug() << "sdaEdge" << (uint8_t)sdaEdge << "sclEdge" << (uint8_t)sclEdge;
-
 	if ((sdaEdge == edge::rising) && (sclEdge == edge::falling)) // INVALID STATE TRANSITION
 	{
         state = transmissionState::unknown;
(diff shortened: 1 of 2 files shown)


Desktop_Interface/i2cdecoder.cpp
```diff
@@ -13,7 +13,15 @@ i2cDecoder::i2cDecoder(isoBuffer* sda_in, isoBuffer* scl_in, QPlainTextEdit* con
     updateTimer = new QTimer();
     updateTimer->setTimerType(Qt::PreciseTimer);
     updateTimer->start(CONSOLE_UPDATE_TIMER_PERIOD);
-    connect(updateTimer, SIGNAL(timeout()), this, SLOT(updateConsole()));
+    connect(updateTimer, &QTimer::timeout,
+            this, &i2cDecoder::updateConsole);
```
EspoTek: It's not necessarily wrong (at all) to split calls like these along two lines, but I personally don't do it.  Are you able to do single line function definitions and calls just for consistency?

SebastianMestre: Ah, my bad. I've been looking throught the Qt docs recently and that's the style they use, I guess it rubbed off onto me.

I'll put it back into a single line.

SebastianMestre:
No problem! Refactoring code that is already clean is not difficult though.

I didn't really mind looking silly but my girlfriend saw the previous picture and thought I looked like I was balding, so she urged me to change it. I don't have many pictures that would be appropriate for a github profile (most I have are with friends, or with my gf, or doing a silly face) so I went with this one, which is a bit of an older one.

EspoTek:
Oh man, you should see my Slack profile picture at work.  A full 25% of the image is taken up by my receding hairline.  Sometimes you just have to embrace what God gives you.  :P

PR #81, opened 2019-05-10
Isodriver refactor - part 1

SebastianMestre:
In this PR, I started moving everything that's related to graphing in isoDriver into its own class, DisplayControl (working name).

For the sake of reasonable PR sizes I am doing it now, even though it's mostly incomplete. There are still a bunch of variables (marked with a TODO comment) that I want to move over, as well as a bunch of methods. Right now I'm passing a whole bunch of things around to make everything work "as before".

The quotation marks around "as before" are there because it's not really working quite like it used to: `updateTimeWindow` is called even if the display is `properlyPaused()`. Before, this was not so. I assumed this was a mistake since i saw no reason not to call it.

Code is pretty messy right now but it should settle down once I'm done with the changes I intend on doing probably by next PR.

I also did slip a few cheeky purely formatting-based changes in, just to make old code fit the style of the more recent code. I tried to make sure that I only give this treatment to code that was going to show up in the diff anyways, as to not pollute the blame.

Speaking of formatting, I've been using clang-format to do these formatting changes automatically. For that purpose, I wrote a really barebones .clang-format file which I think fits the style guidelines you described to me sometime in the past (can't remember if it was on an email or a PR comment) and was wondering if you would be interested in me adding a PR for that?

Alright. it's pretty late and I feel like my writing is all over the place, so I'm gonna wrap this up by asking what's up with all the redundant type casts? Not that it bothers me, I was just wondering...

Anyways, it's nice to be back, warm regards,

Sebastian

### DIFF
--- a/Desktop_Interface/isodriver.cpp
+++ b/Desktop_Interface/isodriver.cpp
@@ -283,7 +283,7 @@ void isoDriver::setVoltageRange(QWheelEvent* event)
     }
     else
     {
-        // Not sure if the following code changes this so I'm storing it here, it will be used again shortly
+        // Not sure if the following code changes the outcome of this call
         const bool isProperlyPaused = properlyPaused();
 
         double c = (display.window) / (double)200;
@@ -293,11 +293,6 @@ void isoDriver::setVoltageRange(QWheelEvent* event)
 
         pixPct /= isProperlyPaused ? (double)(range.upper - range.lower)
                                    : (double)(display.window);
-
-        if(not isProperlyPaused)
-        {
-            qDebug() << "TIGGERED";
-        }
 
         if (pixPct < 0)
             pixPct = 0;
@@ -310,6 +305,7 @@ void isoDriver::setVoltageRange(QWheelEvent* event)
 
         if (not isProperlyPaused)
         {
+            qDebug() << "TIGGERED";
             qDebug() << "upper = " << range.upper << "lower = " << range.lower;
             qDebug() << "window = " << display.window;
             qDebug() << c* ((double)pixPct);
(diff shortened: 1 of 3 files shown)


Desktop_Interface/isodriver.cpp
```diff
@@ -255,125 +248,113 @@ void isoDriver::setVisible_CH2(bool visible){
     axes->graph(1)->setVisible(visible);
 }
 
-void isoDriver::setVoltageRange(QWheelEvent *event){
+void isoDriver::setVoltageRange(QWheelEvent* event)
+{
     if(doNotTouchGraph && !fileModeEnabled) return;
 
+    bool isProperlyPaused = properlyPaused();
```
EspoTek: WIth a single pause state, the `properlyPaused` method can be eliminated and just m_paused can be used there.
Keep it as is for the moment, though.

Desktop_Interface/isodriver.cpp
```diff
@@ -255,125 +248,113 @@ void isoDriver::setVisible_CH2(bool visible){
...
         double c = (topRange - botRange) / (double)400;
 
         QCPRange range = axes->yAxis->range();
 
-        double pixPct = (double)100 - ((double)100 * (((double)axes->yAxis->pixelToCoord(event->y())-range.lower) / (double)(range.upper - range.lower)));
-        if (pixPct<0) pixPct = 0;
-        if (pixPct>100) pixPct = 100;
-
+        double pixPct = (double)100 - ((double)100 * (((double)axes->yAxis->pixelToCoord(event->y())-range.lower) / range.size()));
+        if (pixPct < 0) pixPct = 0;
```
EspoTek: Would prefer separate lines for if() statements, if possible.  Please don't repeat my mistakes!

Desktop_Interface/isodriver.cpp
```diff
@@ -1148,25 +1127,28 @@ void isoDriver::slowTimerTick(){
     update_CH2 = true;
 }
 
-void isoDriver::setTopRange(double newTop){
-    topRange = newTop;
-    topRangeUpdated(topRange);
+void isoDriver::setTopRange(double newTop)
+{
+    // NOTE: Should this be clamped to 20?
```
EspoTek: An assert here would be more appropriate.  If the newTop is > 20, something has gone horribly horribly wrong.

EspoTek:
Welcome back, man!

Looks like a good start.  I noticed a lot of it is formatting changes, though.
If possible, could you do that in a separate review?
Just makes it much easier to read.  Wouldn't be opposed to running `clang-format` recursively on each file in the repo if that would reduce the amount of style edits in future reviews.
I'm not too worried about polluting `git blame`.  Anything that gets committed has been reviewed by myself (if not written), so I should (in theory) know what it does.

Those reduntant casts should bother you - they're very smelly!  I think it's a relic of a younger me that doesn't understand C++ arithmetic rules making sure everything's double just in case...

Desktop_Interface/isodriver.cpp
EspoTek: Actually, I think formatting stuff should go into a single giant review.  :)  See the `clang-format` review.

PR #82, opened 2019-05-11
Add .clang-format file

SebastianMestre:
I wrote a fairly barebones .clang-format file. I've been using it myself to keep my code in the right style.

I figured it might be helpful for other potential contributors.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,5 +1,8 @@
 BasedOnStyle: WebKit
 
+IndentWidth: 4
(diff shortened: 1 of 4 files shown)


EspoTek:
It's a good start.

I can think of a few things off of the top of my head that could be added to improve this further, though:
```IndentWidth: 4
UseTab: Always
AllowShortIfStatementsOnASingleLine: false```

I'm sure we'll find other things as we go along.

PR #103, opened 2019-10-07
Enable smooth zoom

andrewdotn:
The current code interprets wheel events with a scroll distance of 120 units as a zoom in one direction, and all other wheel events as a zoom in the other direction. When a mac touchpad sends many high-resolution events of less than 120 units, the result is that you instantly get zoomed all the way out with no way to zoom back in.

This PR fixes that: trackpad scroll gestures, in both the horizontal and vertical directions, smoothly zoom the time and voltage axes.


EspoTek:
Thanks for this.  Makes sense in theory and it sounds like you've tested it too (which I'm unfortunately unable to do as I don't have a MacBook).  Code is a lot cleaner too.

PR #115, opened 2020-02-17
Allow PSU abort from PSU calibration

jrwwallis:
Hi Chris,

This fixes a couple of minor nits with the calibration UI:

1) If "Calibrate" is chosen from the "Oscilloscope" menu, then there is a check if USB is connected, and calibration doesn't proceed if not connected.  However if "Calibrate" is chosen from the "Power Supply" menu, there is no such check.  An equivalent check is added for UI consistency.

2) After oscilloscope calibration is done, the UI flow goes straight into PSU calibration, whether or not the user wants to do that.  This adds a "Cancel" button to the PSU calibration dialog to allow PSU calibration to be skipped after oscilloscope calibration is complete.

Thanks,

John

### DIFF
--- a/Desktop_Interface/mainwindow.cpp
+++ b/Desktop_Interface/mainwindow.cpp
@@ -2203,6 +2203,13 @@ void MainWindow::on_actionCalibrate_2_triggered()
 {
     qDebug() << "Calibrating PSU!";
 
+    if (!ui->controller_iso->driver->connected) {
+        calibrationMessages->setStandardButtons(QMessageBox::Ok);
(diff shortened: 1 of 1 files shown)


EspoTek:
Thanks for this, John. Makes a lot of sense and solves a real problem in a simple way.
I love the user message, too.  Fits in thematically with all of the others.

jrwwallis:
Thanks Chris.  Actually I just copied the message from https://github.com/EspoTek/Labrador/blob/master/Desktop_Interface/mainwindow.cpp#L1722 :-)
John

PR #162, opened 2021-03-20
Exporting the current oscilloscope view as image

landswellsong:
Hello. Got a quick PR for you, not sure if the code is optimal but would be nice if you had a look.

This adds an "Export Current view" option to the file menu, which captures the current scope view as an image, using the current window dimensions for the picture size.

The primary motivation was to be able to take vector (PDF, but still) screenshots.

[out.pdf](https://github.com/EspoTek/Labrador/files/6175781/out.pdf)

### DIFF
--- a/Desktop_Interface/mainwindow.cpp
+++ b/Desktop_Interface/mainwindow.cpp
@@ -1966,6 +1966,47 @@ void MainWindow::openFileDialog(QString *fileName){
     *(fileName) = temp;
 }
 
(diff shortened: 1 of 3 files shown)


EspoTek:
Thanks so much for your PR, Oleksiy.  I had no idea that you could do this with QCustomPlot.
I've merged your commit into master but with one tiny change (capitalised text in the menu item).

I also checked out your file in Xpdf at 800%.  It's definitely vectorising properly!

Thanks again,
~Chris

PR #166, opened 2021-03-28
Android build 20210328

superfrink:
The new show range dialog setting feature did not built on Android.  This change takes the setting out of the Android build.

At first I was going to the add the setting to the Android UI but the range dialog itself is not in the Android build so the checkbox wouldn't be useful.


EspoTek:
Thank you again!  Not sure why it wouldn't build, but you're right - it's not part of the Android UI anyway.

PR #169, opened 2021-04-03
Don't strip all optimization flags.

sandsmark:
Some basic compiler functionality is dependendent on having
optimizations enabled.

There shouldn't be any issues with -O2, which is the default for qmake's
release builds.

### DIFF
--- a/Desktop_Interface/Labrador.pro
+++ b/Desktop_Interface/Labrador.pro
@@ -251,29 +251,6 @@ unix:SOURCES += unixusbdriver.cpp
 unix:HEADERS += unixusbdriver.h
 
 #############################################################
-########       SHARED ANDROID/LINUX GCC FLAGS      #########
-###########################################################
-
-unix:!macx: QMAKE_CXXFLAGS_RELEASE -= -O0
-unix:!macx: QMAKE_CXXFLAGS_RELEASE -= -O1
-unix:!macx: QMAKE_CXXFLAGS_RELEASE -= -O2
-unix:!macx: QMAKE_CXXFLAGS_RELEASE -= -O3
-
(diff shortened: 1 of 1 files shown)


sandsmark:
appveyor seems to fail because it tries to push something to github but can't authenticate.

EspoTek:
Yep, AppVeyor requires my credentials.  Don't worry too much about it.

Regarding the flag stripping, though, I remember those lines were added to solve (I think?) some weird hang during the Linux USB setup code about 4 years ago.
Not knowing what exactly those flags do, is there any reason why you think we should remove them?

sandsmark:
I don't remember exactly what, but some of the default compiler options used when packaging on archlinux (and some other distros) depend on some compiler analysis passes that are only done when optimization is turned on. So not the optimizations themselves, it just needs the information the compiler accumulates when optimizing. So it is necessary to package it on archlinux at least.

And without any optimizations at all things got slow on my laptop (and the CPU usage on my desktop machine went up a bit).

I have started on a branch that fixes some compiler warnings, I assume one/some of those might be the original underlying reason.

Though I have been using it on linux for a while now without problems (with optimizations).

sandsmark:
the commit is here as well, so closing this: https://github.com/EspoTek/Labrador/pull/172

mikhirev:
> Regarding the flag stripping, though, I remember those lines were added to solve (I think?) some weird hang during the Linux USB setup code about 4 years ago.

It means there's (or was) some bug in code. It needs to be fixed, removing optimization flags is a wrong way to work around errors.

EspoTek:
Yes, I know that now.  But it doesn't mean I should resurrect some undefined behaviour from 4 years ago just for the sheer intellectual joy of applying a "proper" fix.

PR #177, opened 2021-05-10
license clarification

mikhirev:
- added GPL text
- added copyright notices to source files
- removed trailing spaces

### DIFF
--- a/AVR_Code/USB_BULK_TEST/src/globals.h
+++ b/AVR_Code/USB_BULK_TEST/src/globals.h
@@ -3,7 +3,38 @@
  *
  * Created: 18/04/2015 12:44:42 PM
  *  Author: Esposch
- */ 
+ */
+
+/* Copyright (C) 2018 Christopher Paul Esposito
+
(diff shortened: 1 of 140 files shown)


EspoTek:
Hi mate.  Sorry about this, but I'd prefer not to have copyright info at the top of each file.
If you'd like to make a new PR removing trailing whitespace, I'd be happy to accept that, though.

mikhirev:
> I'd prefer not to have copyright info at the top of each file.

Why? It is a common practice and it is done [for purpose](https://www.gnu.org/licenses/gpl-faq.en.html#NoticeInSourceFile).

Removing trailing spaces is just a side effect of automatic files processing.

EspoTek:
Mainly because it looks ugly.
I'll usually go out of my way to delete auto-generated licences from the top of files I actively work with.

Regarding how common the practice is, it's obviously not uncommon, but it's far from a universal standard.  The first three projects I looked up (Homebrew, Alacritty and stb) all didn't have the licence at the top of each file, although stb did have it at the bottom.

PR #187, opened 2021-10-29
Replace -lomp with -fopenmp

turboencabulator:
On my machine, libomp is part of clang, and libgomp is part of gcc.  Using -fopenmp should let it choose the right (or at least the existing) library based on the compiler.


EspoTek:
Good call!

PR #189, opened 2021-10-30
Fix a few compiler warnings

turboencabulator:
Some low-hanging fruit.


EspoTek:
This is an uphill battle, but a good start nonetheless!  Thanks so much, merging now.

PR #192, opened 2021-11-06
Whitespace cleanup

turboencabulator:


EspoTek:
Always get scared by big whitespace commits.
Here, it looks like you've replaced all the tabs with 4 spaces, including in some files with a CSV-esque format that relies on tabs to delineate (the .tlw files).

Are you sure this won't cause any issues?

turboencabulator:
The only change to the .tlw files was adding a newline at the end-of-file.  2 of the 5 were missing it, so if the other 3 were loading OK then I don't expect any problems from it.

I also added a similar newline to the end of _list.wfl.  It's loading OK for me on Linux, I think at worst you'd get an extra empty item at the end of the combo box if the parser gets confused by it (and selecting that empty item would crash it).  BTW, I'm brewing up some changes to simplify the parsers, but wanted to get the whitespace fixes in first so there aren't any merge conflicts.

Then there are the two .bat files where I also added the newline but I can't verify if they work.  It may need a line in a .gitattributes file to convert their line endings to CRLF on checkout in case Windows gets confused (since previously there was only one line and no newline characters at all).  https://serverfault.com/questions/429594/is-it-safe-to-write-batch-files-with-unix-line-endings seems to suggest any line ending problems in batch files are limited to gotos and labels, so I don't expect any problems here.

EspoTek:
OK.  As long as it's tested, and it serves a useful purpose (avoiding merge conflicts is always nice!) I'm happy to merge.

PR #194, opened 2021-11-09
Simplify the waveform reading code

turboencabulator:
Combine the Android and non-Android implementations.  Only difference
now is the path to the waveform data.

Remove arbitrarily-sized buffers and most C-style string handling
function calls.  Use native Qt string methods instead; this should
improve memory safety.


EspoTek:
Thanks!  So much more compact now.

PR #196, opened 2021-11-14
Resize icon and reduce image file sizes

turboencabulator:
The 256x256 icon was actually 48x48.  Regenerate it using the PCB artwork.  Also do a couple other conversions to reduce file sizes.


mi-hol:
@turboencabulator is this still applicable?
In case it is, please go ahead with merging

turboencabulator:
Yes, the wrong icon size causes lintian to complain about the Debian package.  Merging it now.

PR #210, opened 2022-01-28
Make Librador logger configurable

couchand:
This adds a layer of indirection to logging in Librador.  Rather than directly calling `(f)printf`, we call through a global logger that the user can override.

The basic structure of this logger is taken from similar code in simavr: https://github.com/buserror/simavr/blob/ea4c4504d15117223a23e2dd6edb745fea61ceae/simavr/sim/sim_avr.c#L40-L73


EspoTek:
Ship it!

PR #235, opened 2023-01-18
update Multimeter UI

jimustafa:
The Multimeter section in the wiki refers to a reference resistor... Did I get this right?


EspoTek:
You're right, this wording is clearer.  Thanks for the patch.

PR #263, opened 2024-03-23
Serial encoding support

mmehari:
In this pull request, I will like to bring serial encoding support for Espotek/Labrador.

This topic has been discussed in a previous issue (https://github.com/EspoTek/Labrador/issues/239) but instead of using GPIOs, I implemented it using the Signal Generator on CH1.

As Signal Generators make use of the hardware DMA support for playing waveforms repetitively, I extend this functionality for playing one-time waveforms. This way, a character pressed on the keyboard is sent to the hardware for it to be played once.

User interface wise, a single checkbox ("Serial Encoding") appears under Logic Analyzer CH1 and clicking it opens a new (one row) text-box. Everything written here will be serially transmitted over Signal Gen CH1. It also hides waveform, frequency and offset input parameters on Sig Gen CH1 since they are not used for serial communication. 

I also extended the vendor request commands for sending the serial data to the hardware over USB.

Finally, I also added a filter for ANSI escape codes. After successfully controlling a raspberry pi4 hardware, I came to see ANSI escape codes being received and they were not been interpreted at our side. I then decided to filter them out before reaching the display buffer.

**Note**:
Tested on raspberry pi5 hardware + rpi OS


EspoTek:
Wow, this looks really solid!  The only complaint I could possibly make is that it too perfectly integrates into my own ugly code.  :P

Before I merge this into `master`, are you quickly able to confirm the changes to the .hex files?

Specifically:
 - `labrafirm_0007_01.hex` should have been compiled with  `#define SINGLE_ENDPOINT_INTERFACE` commented out
 - `labrafirm_0007_02.hex` should have been compiled with `#define SINGLE_ENDPOINT_INTERFACE` enabled
 - Both firmwares should have been compiled as debug (or release with `-O0`), with `#define OVERCLOCK 48` enabled

If you're unsure I can quickly recompile from here before this patch goes through.

Thanks again for sharing your work with everyone!

Cheers,
~Chris

mmehari:
Indeed that is correct.
`labrafirm_0007_01.hex` is compiled for **non** `SINGLE_ENDPOINT_INTERFACE` and `labrafirm_0007_02.hex` is compiled for `SINGLE_ENDPOINT_INTERFACE`.

Also `#define OVERCLOCK 48` is enabled during both firmware compilations.

However, I wasn't paying attention on the (release) `optimization flag` and it was set to -Os after checking it again.

EspoTek:
Perfect!  I'll test on Windows soon and then push the patch.  Thanks again!

mi-hol:
> Perfect! I'll test on Windows soon and then push the patch. Thanks again!

Would perhaps the changes in Makefile.Debug & Makefile.Release of commit 42a2b9763dcd5e1ce7d43b41db2f97a178c2d862 have addressed remaining open actions?

EspoTek:
> > Perfect! I'll test on Windows soon and then push the patch. Thanks again!
> 
> Would perhaps the changes in Makefile.Debug & Makefile.Release of commit [42a2b97](https://github.com/EspoTek/Labrador/commit/42a2b9763dcd5e1ce7d43b41db2f97a178c2d862) have addressed remaining open actions?

Nope, I think the patch was actually good the whole time.  Just started a second business and had a baby.  Things have been nuts.  Going to merge it now!

mi-hol:
Thanks for merging, congrats to having a baby and good look for the new business

PR #270, opened 2025-01-13
Introducing duty cycle to the Square waveform

mmehari:
This pull request adds duty cycle on the square waveform of Signal Gen CH1 and CH2.

Under the hood, the locally stored waveform data (**Square.tlw**)  is replaced by a  **duty-cycled** Square waveform and sent over to the board for waveform generation.

The duty cycle takes percentage values (0% - 100%)


EspoTek:
Thanks mate, just checked and merged!

PR #272, opened 2025-01-31
Frequency response calculation

mmehari:
This PR will bring **frequency response calculation** to espotek's oscilloscope functionality.
Interface wise, it will be located under the oscilloscope tab, next to _Frequency Spectrum_.

freqResp_interface

Operation wise, a frequency response calculation requires
1. input source to excite the circuit-under-test (CUT) with sinusoidal waveform of varying frequencies, and
2. two output measurements for capturing the excitation and response wave-forms

To this end, Signal Gen CH1 will be used for 1 and Oscilloscope CH1/CH2 will be used for 2.

Here is an example circuit-under-test (CUT) for measuring the frequency response of an RC circuit.

freqResp_CUT

### Frequency response calculation steps
1. generate a sinusoidal frequency tone on Signal Gen CH1
2. capture 1 cycle of input and output wave-forms
3. using least squares regression, fit a sinusoidal to captured wave-forms
4. calculate gain(dB) and phase response (degree)
5. plot frequency response (gain or phase)
6. update next frequency tone on Signal Gen CH1 and start all over again

freqResp_output

**Note**:
Frequency response calculation has been tested on Linux and RPI5 platforms.

### DIFF
--- a/Desktop_Interface/isodriver.cpp
+++ b/Desktop_Interface/isodriver.cpp
@@ -1,3 +1,4 @@
+#include <Eigen/Dense>
 #include "isodriver.h"
 #include "isobuffer.h"
 #include "isobuffer_file.h"
@@ -7,6 +8,7 @@
 #include <iostream>
 #include <omp.h>
 
+#define PI 3.141592653589793  // Predefined value for pi
 static constexpr int kSpectrumCounterMax = 4;
 
 isoDriver::isoDriver(QWidget *parent) : QLabel(parent)
@@ -617,7 +619,6 @@ void isoDriver::setTriggerMode(int newMode)
 //0 for off, 1 for ana, 2 for dig, -1 for ana750, -2 for file
 void isoDriver::frameActionGeneric(char CH1_mode, char CH2_mode)  
 {
-
     // The Spectrum is computationally expensive to calculate, so we don't want to do it on every frame
     static int spectrumCounter = 0;
     if(spectrum)
@@ -631,6 +632,9 @@ void isoDriver::frameActionGeneric(char CH1_mode, char CH2_mode)
     internalBuffer375_CH1->enableDftWrite(spectrum);
     internalBuffer375_CH2->enableDftWrite(spectrum);
     // internalBuffer750->enableDftWrite(spectrum);
+
+    internalBuffer375_CH1->enableFreqResp(freqResp, freqValue_CH1->value());
+    internalBuffer375_CH2->enableFreqResp(freqResp, freqValue_CH1->value());
 
     //qDebug() << "made it to frameActionGeneric";
     if(!paused_CH1 && CH1_mode == - 1){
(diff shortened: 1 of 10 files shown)


MatthewSteeves:
This is awesome! Great to see these enhancements!

mmehari:
That is correct, and I noticed it after making the PR.

mi-hol:
Would you mind to edit the text, so it could be directly copied to WIKI page as documentation, after merge of PR?

mmehari:
I have made the change

mi-hol:
@turboencabulator would perhaps reviewing this PR work for you?
In that case please assign yourself as "reviewer"

Desktop_Interface/mainwindow.cpp
```diff
@@ -280,6 +285,64 @@ MainWindow::MainWindow(QWidget *parent) :
...
+    QLabel* freqRespMaxFreqLabel = new QLabel("Max Frequency (Hz)");
+    QLabel* freqRespStepLabel = new QLabel("Step (Hz)");
+    QLabel* freqRespTypeLabel = new QLabel("Response");
+
+    freqRespLayout1Widget->setLayout(freqRespLayout1);
+    freqRespMinXSpinbox->setMinimum(100);
+    freqRespMinXSpinbox->setMaximum(62500);
+    freqRespMaxXSpinbox->setMinimum(100);
+    freqRespMaxXSpinbox->setMaximum(62500);
+    freqRespMaxXSpinbox->setValue(32500);
```
turboencabulator, line 308: When I enable this mode, there is a mismatch between what this spinbox shows (32500) and the actual value it's using (62500, from `m_freqRespMax`).  It steps right past 32500.  I think the other spinboxes are behaving similarly.

Desktop_Interface/isodriver.cpp
```diff
@@ -821,6 +856,127 @@ void isoDriver::frameActionGeneric(char CH1_mode, char CH2_mode)
...
+                        m_freqRespFreq.clear();
+                        m_freqRespGain.clear();
+                        m_freqRespPhase.clear();
+                        m_freqRespFlag = false;
+                    }
+
+                    // Select new frequency
+                    double freqValue = freqValue_CH1->value() + m_freqRespStep;
+                    if(freqValue >= m_freqRespMax || freqValue < m_freqRespMin || m_freqRespFreq.size() == 0)
+                        freqValue = m_freqRespMin + m_freqRespStep;
```
turboencabulator, line 954: Should the endpoints be inclusive?  `freqValue > m_freqRespMax` instead of `>=`, and omit the `+ m_freqRespStep` on the reset.

Beware my other comment though about the spinboxes not showing the values it's actually using.  I tried it and I think it was using `m_freqRespMin` of 0 (or very close to it) instead of the 100 that the spinbox shows, and setting the frequency that low made it hang.

turboencabulator:
@mmehari Awesome work!  Do you know if there is a maximum frequency that this supports?  I built the same circuit in your photo and it only plots the line to about 2.4kHz even though it steps past that.  Not sure if I'm doing something wrong or if it's just a limitation somewhere.

Desktop_Interface/mainwindow.cpp
mmehari: Indeed that is correct.
The spinboxes (freqRespMinXSpinbox, freqRespMaxXSpinbox) and their shadow registers (m_freqRespMin, m_freqRespMax) are not set to similar default values and for the case when the spinbox values were not changed, the current frequency steps right past it.

I have issued a bug fix for this (https://github.com/espotek-org/Labrador/pull/272/commits/e223e1c891e067c5cd0403f2a1bbbac9403617cb)

Desktop_Interface/isodriver.cpp
mmehari: This is also correct and the new frequency value selection should use non inclusive endpoint value. A bug fix is also submitted for this (https://github.com/espotek-org/Labrador/pull/272/commits/7c8f42f3e0f221caf081be2d334efdb94bfd270b).

The reason the frequency response calculation hangs at lower frequency values is because it takes very long to collect samples for one complete cycle. But now, it shouldn't happen anymore as the default value for both the spinbox and the shadow register is set to 100Hz.

mmehari:
> @mmehari Awesome work! Do you know if there is a maximum frequency that this supports? 

Well the maximum frequency supported is related to the range of Signal Gen CH1, which is 62.5KHz.

>I built the same circuit in your photo and it only plots the line to about 2.4kHz even though it steps past that. Not sure if I'm doing something wrong or if it's just a limitation somewhere.

It depends on a couple of things
1. You might have used different RC values. Different Time Constant (τ) results in different frequency response.
2. Lower input voltage results in lower output voltage, thus will not be able to plot higher frequency contents. Try using higher input voltages (9V) and see if you get past 2.4KHz

_Note_: it is possible that the frequency response plot jumps steps without plotting them. The frequency response calculation relies on least square fitting to collected samples and if the samples have lower magnitudes, the fitting residual (error) becomes very large and the step gets discarded.

mi-hol:
@mmehari @turboencabulator 
Great conversation, resulting in further improvements!
This is collaboration at its best! 

> _Note_: it is possible that the frequency response plot jumps steps without plotting them. The frequency response calculation relies on least square fitting to collected samples and if the samples have lower magnitudes, the fitting residual (error) becomes very large and the step gets discarded.


Very good explanation for real world issues! 
Would it be fair to say that newbies (like me) would usually be unaware of the complexities of "what is going on behind the scenes" and consider "the step gets discarded" (I assume without a visual indicator of the fact) as an unexpected behaviour?

mmehari:
It is not an unexpected behaviour in the sense of a bug from the calculation, but it is more like a limitation of the applied methodology. In the course of the frequency response calculation, you will encounter frequency values where the response from the circuit-under-test (CUT) is very weak (higher frequencies in RC circuits) and that can't be used to evaluate the frequency response. Thus, you simply discard it and continue on to the next frequency value.

mi-hol:
> it is more like a limitation of the applied methodology

Thanks for clarification

mmehari:
Along the way, I found one extra bug.
The thing is oscilloscope CH2 UI is disabled in frequency response mode and it was not re-enabled when switching to frequency spectrum mode. The commit https://github.com/espotek-org/Labrador/pull/272/commits/a90deb6b3675baab9ec9ddb46539a912cd851470 will fix this.

turboencabulator:
Thanks @mmehari!  I think I've figured out why it wasn't plotting.  The signal generator clips all of the sine wave peaks at 3.3V, and the valleys start to distort when the frequency goes past ~400Hz.  (The falling edge slew rate looks like it's limited when below ~0.9V, probably a hardware limitation?)  All this distortion introduced too much error, causing the measurement at that frequency to be discarded.  I got the best results by setting the offset to 1V and amplitude to 2.2-2.3V.  Results look good out to about 6kHz and then they start to get noisy and unreliable past 10kHz.

I've got a couple of wishlist items, I wouldn't consider any of these to be blockers though:
* A button to clear the display and restart the measurement.  I was clicking the up/down arrows on the min/max/step spinboxes to get it to restart, and it would be nice if there was a dedicated button for it.
* Speaking of the up/down arrows, it would be nice if they incremented the most-significant digits instead of incrementing/decrementing by 1, like the frequency/amplitude/offset spinboxes.

mmehari:
I have addressed both wishlists and let me know if that is what you have asked

turboencabulator:
That's exactly what I was looking for.  Thanks!

mi-hol:
> That's exactly what I was looking for.

@turboencabulator 
image

Marking all comments as "resolved" via button 
image

would complete the review "officially" and open the merge window.

Example of open conversation:
image

mi-hol:
> This is awesome! Great to see these enhancements!

@MatthewSteeves would you perhaps mind to become the beta tester?

EspoTek:
First time properly looking through this thread.  This is awesome man.

turboencabulator:
> would complete the review "officially" and open the merge window.

I think I clicked them all.  (First time using the review feature.)  Let me know if I missed any.

mmehari:
The frequency response calculation requires both Oscilloscopes (CH1 and CH2) to be present and receive samples.
Hence it beneficial to make them non checkable. commit https://github.com/espotek-org/Labrador/pull/272/commits/84541a90dde430e92ac3408eaa25d461aec140b0 does this.

mi-hol:
> First time using the review feature

with success! 

@mmehari  LGTM (from list of code-review acronyms at  https://www.codelantis.com/blog/code-review-acronyms-lgtm-nit)

Please go ahead and merge code, then update wiki

MatthewSteeves:
> > This is awesome! Great to see these enhancements!
> 
> @MatthewSteeves would you perhaps mind to become the beta tester?

Hey I’ll take a look and provide feedback, though I’m not very qualified to really put the change through it’s paces. My background is IT/programming and electronics of (one of) my (too many) interests! 😄

mmehari:
Wiki is updated.
https://github.com/espotek-org/Labrador/wiki/Frequency-Response-Calculation

PR #290, opened 2025-02-16
Repair AppImage build

turboencabulator:
`make_appimage` has begun to rot, it depends on one download that no longer exists, and I get a few errors running it locally.  This set of changes lets me at least build the package, but I can't vouch for its portability yet.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -29,7 +29,6 @@ jobs:
       - name: Run uploadtool for Linux
         run: |
           export UPLOADTOOL_SUFFIX="Linux"
-          find appdir -executable -type f -exec ldd {} \; | grep " => /usr" | cut -d " " -f 2-3 | sort | uniq
           export UPLOADTOOL_BODY="EspoTek Labrador software for Linux. This should be compatible with all modern 64-bit distros. If you're running 32-bit Linux, please see the wiki for build instructions. https://github.com/espotek-org/Labrador/wiki/Building-for-32-bit-Linux"
           wget -c https://github.com/probonopd/uploadtool/raw/master/upload.sh
           bash upload.sh EspoTek_Labrador*.AppImage
--- a/.travis.yml
+++ b/.travis.yml
@@ -87,7 +87,6 @@ after_success:
     # Linux
     elif [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
       export UPLOADTOOL_SUFFIX="Linux"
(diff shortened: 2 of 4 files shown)


mi-hol:
@turboencabulator thanks for that improvement!

> but I can't vouch for its portability yet

Would you mind to elaborate on what can we can do to assist?

turboencabulator:
The most important thing would be to finish setting up the automated builds.  Here's the situation as I understand it:
* `linuxdeployqt` copies most of the library dependencies from the build machine's environment into the package it creates.
* Its guidance is to run this on the oldest supported machine to both make sure it builds (doesn't depend on anything too new) and so that it copies older versions of libraries (with the assumption they will continue to work on newer environments).  The workflow stuff is currently set up to run on Ubuntu 20.04 instead of latest, this being the oldest version that the workflows currently support.
* When I say "most" of the libraries, it's not packaging the very lowest level ones such as `libstdc++`, so the above point becomes important.  Otherwise some of the packaged code might depend on a newer `libstdc++` or other lib than what your machine has.
* We're copying `libstdc++` from the build machine into the package manually, along with another library whose download is now a dead link, and doing some fixups to use them.  I assume this mostly solves that dependency problem, but now without one of the two libraries then maybe these fixups no longer work correctly.
* It looks like the reason the download went away is because that project is now a plugin for a different but similarly named program called `linuxdeploy`.  Maybe we need to switch over to that.

EspoTek:
I am 99% sure that the .so file from darealshinji did something important- I just can't quite remember what (I think it's related to compatibility on older Linux versions, but not 100% certain).

For the sake of Chesterton's fence, should we maybe just upload the .so file to the repo and copy it into the /usr/optional directory from there?

If you run the AppImage from GitHub Releases with the argument --appimage-extract, it extracts the full (virtual) filesystem, including exec.so:
https://github.com/espotek-org/Labrador/releases/download/continuous-Linux/EspoTek_Labrador-2cc0678-x86_64.AppImage

turboencabulator:
Looking into it more, the missing file `exec.so` is a replacement for `execve` and friends, to keep the AppImage environment from leaking into child processes.  It's a two-line change to build our own copy of it, I'll have a PR to do it shortly.  I much prefer that over committing mystery binary blobs or 3rd-party sources to the repo.

As a long-term fix, we should probably switch to using darealshinji's `linuxdeploy` plugin as intended in case anything changes again.  But before we can do that, we need to factor out the udev rule installation from our custom AppRun script and write our own plugin to reinsert it.  More PRs to come.

EspoTek:
100% agree - your solution is better!

PR #291, opened 2025-02-18
Restore exec.so to the AppImage

turboencabulator:
Build it from source instead of trying to fetch a prebuilt copy.

Fixes lingering issues from #290.

### DIFF
--- a/Desktop_Interface/make_appimage
+++ b/Desktop_Interface/make_appimage
@@ -5,13 +5,18 @@ set -e
 qmake APPIMAGE=1 CONFIG+=release PREFIX=/usr DEFINES+=GIT_HASH_SHORT=$(git rev-parse --short HEAD)
 make -j$(nproc) ${CC:+CC=${CC}} ${CXX:+CXX=${CXX} LINK=${CXX}}
 make INSTALL_ROOT=AppDir install ; find AppDir/
+
 wget -N -nv "https://github.com/probonopd/linuxdeployqt/releases/download/6/linuxdeployqt-6-x86_64.AppImage"
 chmod a+x linuxdeployqt-*.AppImage
 unset QTDIR; unset QT_PLUGIN_PATH ; unset LD_LIBRARY_PATH
 export VERSION=$(git rev-parse --short HEAD)
 ./linuxdeployqt-*.AppImage AppDir/usr/share/applications/*.desktop -bundle-non-qt-libs -verbose=2 -no-strip -unsupported-allow-new-glibc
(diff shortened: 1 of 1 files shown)


EspoTek:
This is honestly probably being overly anal, but I've forked darealshinji's repo to https://github.com/EspoTek/linuxdeploy-plugin-checkrt

Can you please change the `wget` to source from there instead?
This way if they make any breaking changes in future, it won't break us.

I remember this being an issue with another repo (`raspberry-pi-gcc-binary`).

turboencabulator:
Done.

But on the other hand, if we use your fork as an actual plugin, we have to build it into a script first.  Give me a couple days and I think I can get everything set up to use this plugin architecture, and I'll let you decide what works best.

EspoTek:
My fork should be identical to darealshinji's.  Is it not?

turboencabulator:
It doesn't have the equivalent of https://github.com/darealshinji/linuxdeploy-plugin-checkrt/releases/tag/continuous

Maybe that's a button click away?  I don't know enough about generating releases yet.

mi-hol:
> Give me a couple days and I think I can get everything set up to use this plugin architecture, and I'll let you decide what works best.

changing the PR to "draft" would buy that time without the risk of anybody unaware of this task to merge the PR
Link "convert to draft" is at top right side in "Reviewers"
image

EspoTek:
> It doesn't have the equivalent of https://github.com/darealshinji/linuxdeploy-plugin-checkrt/releases/tag/continuous
> 
> Maybe that's a button click away? I don't know enough about generating releases yet.

Gotcha.  If it's not necessary to build the AppImage, though, I think we can just ship this as is and deal with that problem should it arrive in the future.  I can always manually add it to releases should that be an issue.

mi-hol:
> changing the PR to "draft" would buy that time, without running the risk of anybody unaware of this task, to merge the PR.

@turboencabulator with #290 merged, will this PR be required still?

turboencabulator:
It's still an open question of whether we want to do this or #293.

mi-hol:
Sound decisions are from my view only possible when the pros & cons of each alternative are known.
Unfortunately I don't have that insight.
Who would be in a position to make a recommendation?

mmehari:
I am not the right person for this either. May be @EspoTek is the right candidate

EspoTek:
I've shared my opinion/experience in #293 - I think this option, while arguably less elegant, is more stable long-term.

turboencabulator:
Yes.  In short, a Github "release" has references to both a git commit hash and a git tag, and that tag doesn't necessarily have to point at the same commit hash.  That's what is going on here, the release points to a new commit, but it also points to a tag that points to an old commit.  It's dumb and confusing, and I don't know why the Github people designed it that way.

I've been testing some alternatives to the current upload script and I haven't yet found any that do exactly what I want (delete the old tag, update the release assets, update all the references).  I'm also experimenting with writing my own upload script.

PR #292, opened 2025-02-18
udev rules cleanup

turboencabulator:
Updated:
* changed the bullet list to numbered list in order to allow for easy and precise feedback (following DRY principle)

1.  Remove the extra copy of the rules in the `AppRun` script so we don't have two copies in the repo that can get out-of-sync.

2. There are three locations where the rules can be placed, `/etc`, `/lib`, and `/run`.  We're currently using `/etc`.  If we're building a package, they should go into `/lib` (which may be the same as `/usr/lib` on systems that did the `/usr` merge), leaving `/etc` as a place where they can be copied to and modified by the user as desired.  If you currently have a copy in `/etc`, I recommend deleting it.  Also see next bullet.

3. When the AppImage runs, it now places the rules in `/run` instead of `/etc` since this location is volatile and we don't want to leave around config files on the user's machine.  Especially if we change the rules in the future and the old stale copy in `/etc` has priority.  My `/run` is cleared after a reboot.  As a side effect, you'll see a `sudo` prompt more often when running the AppImage (whenever the copy in `/run` is cleared).

4. The AppImage now reloads the rules after copying them into place.  Previously udev may not have noticed the new file, and you wouldn't get permission to access the USB device until it did.

5. `udevadm trigger` replays all events (which causes some nuisances on my machine), but what we really want is to replay any for our particular USB device if it's already plugged in.  I've added `--subsystem-match=usb` to filter it down to just the USB events for now; I'm not sure of the right way to make it more specific.

6. If the device is in bootloader mode, the rules created three symlinks (`ATXMEGA32A4U`, `DFU`, and `Bootloader`) instead of one because spaces aren't valid in symlink names.  I've replaced spaces with underscores.  I doubt that the symlinks are useful (and I think having multiple devices plugged in at once might cause confusion here), but I've left them in place.


mi-hol:
Again, a great explanation allowing low skilled people like me to understand (at least I hope so)!

re 1. LGTM

re 2. "we're building a package, they should go into /lib" LGTM

re 3. & 4. LGTM

re 5. sorry, I have no clue

re 6. " I doubt that the symlinks are useful " , in case others share the doubt, why not remove them?

PR #293, opened 2025-02-19
Switch to linuxdeploy + plugins to build AppImage

turboencabulator:
Here's the alternative to #291.  The basic idea is to split our custom `AppRun` into two parts:  The stuff no longer necessary because it is handled by the linuxdeploy `checkrt` plugin, and the udev rules helper.  The new `linuxdeploy-plugin-udev.sh` script handles the latter.

This includes a few small cleanups.  If we go with #291 instead, these should be ported over:

1. Delete `AppDir` and recreate it, to avoid any stale files getting built into the AppImage.
2. Print a message just before the `sudo` prompt explaining that it's installing udev rules, in case anyone wonders what it's attempting to do.
3. For security reasons, use `mktemp` to create the helper script, and don't give everyone execute permission.  Otherwise the script you're about to `sudo` might be writable by others.

As for the differences between `linuxdeployqt` and `linuxdeploy`, I notice it's packaging most of the same libraries but there are a few added and a few removed.  I don't know why.  Also I don't know if we still need the `undef`s in `make_appimage`.


mi-hol:
> Here's the alternative to #291
@turboencabulator what alternative do you recommend?

turboencabulator:
I like this one in that it gets rid of some technical debt, we no longer need to manually insert some libraries and scripts into the AppImage.  If we're going to continue to ship darealshinji's `checkrt` libraries, and that project was made to work nicely with `linuxdeploy`, then I don't see a good reason to try to shoehorn it into `linuxdeployqt` instead if both of those are doing essentially the same thing... unless we have good reason to choose one over the other.  I'll defer to @EspoTek's experience supporting the AppImage on whether there is anything to watch out for.

The drawbacks would be that we're now fetching 3 files from "continuous" builds.  It sounds like @EspoTek might have been burned by that in the past, so it's completely understandable if that's a dealbreaker.  #291 is "stable" in that we're always fetching the same file and/or tool so it won't change on us, but it also means they may become old and unsupported.  We're fetching `linuxdeployqt` version 6 which has been labeled as ["Do not use anymore"](https://github.com/probonopd/linuxdeployqt/releases) (along with all other tagged builds, which I really think is supposed to mean "unsupported").

EspoTek:
I'd definitely prefer to "bake in" outdated versions of libs (whether we build from source or just stuff the executables in) than rely on an external dependency that is constantly changing.  It's less "pretty"/"neat", but more reliable and stable.

That said, nothing wrong with external dependencies if we fork them ourselves and keep everything at a stable version.

turboencabulator:
It looks like two of the three do have tags and should remain stable.  But until we get all three, I'll go ahead and merge #291 along with the cleanups I listed above.  We can revisit this in the future if the situation changes.

turboencabulator:
All the external dependencies are now tagged.  Any other concerns with switching to this AppImage builder?

EspoTek:
> All the external dependencies are now tagged. Any other concerns with switching to this AppImage builder?

When you say they're tagged, are they part of a repository that we "own"?

turboencabulator:
No, but neither is the `linuxdeployqt` file we're currently depending on.

mi-hol:
@turboencabulator sounds like this is ready to merge, right?

turboencabulator:
I have another set of changes brewing that needs to add more to the AppRun script(s) and would conflict, so I'll hold off a bit longer.

turboencabulator:
It turns out that we don't really need a plugin to install scripts, instead we can copy them into the `AppDir/apprun-hooks` directory before running the tool.  This makes it a lot easier to extend.

More benefits from switching:
* Noticeably faster.
* Now supports creating AppImages for ARM and 32-bit x86.
* The Github Actions Ubuntu 20.04 runners are being deprecated and may not be available in the next few weeks, and `linuxdeployqt` won't run as-is without installing some dependencies (`libfuse2` or `libfuse2t64`, maybe others) on newer runners.  No problems with `linuxdeploy`.
* If the old version of `linuxdeployqt` we're currently using (v6) eventually stops working on new runners, I've found that newer versions fail to run unless we manually install `/usr/share/doc/libc6/copyright` into `AppDir`.  Again, no problems with `linuxdeploy`.

I'm going to go ahead and merge this to get ahead of the Ubuntu 20.04 deprecation.

Issue #286, opened 2025-02-13
Doubling sample rate should not suppress the spectrum analyzer


tdunning:
Currently double sample rate apparently suppresses the spectrum analyzer. It would be nice if doubling the sample rate had no effect on the spectrum analyzer other than providing more frequency resolution. This should require no changes other than getting the additional data to the spectrum analyzer and letting the spectrum analyzer know the new sampling frequency.

mmehari:
We can close this issue since the related PR is now merged

Issue #289, opened 2025-02-14
Spectrum display should make use of windowing


tdunning:
When doing a spectral analysis, it is important to multiply each window of input samples by a so-called windowing function. This causes samples at the beginning and end of the window being reduced to zero with the original amplitude being retained in the middle of the window.

The reason that this is important is that the Fourier transform assumes the input is periodic. A batch of samples isn't, however, and there can be considerable mismatch between the beginning of the sample window and the end. That mismatch causes very poor quality in the quality of the display.

As an example, here are a few cycles of a sine wave:

<img width="729" alt="Image" src="https://github.com/user-attachments/assets/cb2d56b9-af48-431f-b357-55e5e9245afc" />

You can see how the first sample on the left is 0 and the last sample is about -1. If you do an analysis that presumes this sampled signal repeats these samples over and over, there will be a big jump from -1 back to 0 when you go from the end back to the beginning.

It is better to use a windowing function. A good choice is $2 \thinspace \sin^2({\pi t} / (n \Delta t))$. This changes our sine wave into this:

<img width="728" alt="Image" src="https://github.com/user-attachments/assets/50d84b04-533f-4e4b-8596-0fe110959ed3" />

Here, the windowing function is shown as a light red ghost on top of the windowed signal. 

In the spectral display, this makes a lot of difference because the step from the end back to the beginning is an impulse that raises the noise floor dramatically.

For instance, here is a plot of the signal above with and without windowing. The difference in noise floor is about 150dB which is massive.

<img width="739" alt="Image" src="https://github.com/user-attachments/assets/4a198c4a-d9c5-4de6-936b-d710aeed3a86" />

In such a spectral displace, the vertical display should be calibrated using a factor based on the number of samples in a window and the gain of the input signal. If we want to display dBm (dB referenced to 1mW), and we have n = 10000 samples in our data and the input gives us a value of 127 for a 1V input, a 1dBm signal would appear to have a magnitude of 109dB after doing the Fourier transform which is computed using

$$s_{ref} = 20 \thinspace \log_{10} (127 \thinspace v_{0dB} \thinspace n)$$

where $v_{0dB} = \sqrt{0.05} = 0.22$ is the voltage that results in 1dB into a $50 \Omega$.

EspoTek:
Wow, that makes a lot of sense.  Thanks for the super clear explanation!

tdunning:
Happy to help as much as I can. Which, sadly, isn't much.

[...]

Issue #63, opened 2019-01-09
Feature Request: FFT / Spectrum Analyzer


drtrigon:
Please add the option to view the data on the GUI in frequency domain also. As far as I understand, that basically only needs post-processing the data with FFTW (or similar). This affects the display/view of the data in the GUI only (no changes to firmware needed).

EspoTek:
Yeah, this is a good idea.  Simple to add and pretty handy.

stuartc:
I would absolutely love this, for diy audio amplifier design and experimentation this would be huge for me!

EspoTek:
Het dit, meneer.  :)

I'll need to look into the specifics of how this is done though, specifically how much data should be used per frame and and how often the display should be updated.  To tell you the truth, I've only ever used spectrum analysers for high frequency wireless, where you don't have to worry about the two time periods being on the same order of magnitude.

PWTim:
I am jusy finishing my trade course for Télécommunications and i avid capable with a SA VNA OTDR equipment and definitely a low spectrum end Even for good reference thats reliable should be a good extension, 8m sure the tiny SA and NanoVNA are great examples on budget friendly but not quality comprising

sandsmark:
fftw is a bit of a hassle to use, I think it would be easier to just pull in this (which also has GPU acceleration): https://github.com/jdupuy/dj_fft/blob/master/dj_fft.h

Vincenzo-Petrolo:
I suppose this issue can be closed now, right?

tdunning:
Yes. It can be closed, I think since the spectrum display is there.

There are still many issues with the spectrum display, but they can reasonably be dealt with in other issues.  For instance,

- double sample rate apparently suppresses the spectrum analyzer
- can't change the frequency scale
- the trigger status display obscures some of the spectral display
- it would be good to do a sin^2 or sin^4 window on the signal to suppress higher frequencies (only useful if you can change the frequency scale)

Even as this is, it is pretty nice to have.

Vincenzo-Petrolo:
Yeah, that's absolutely right. However as you correctly said, another issue should be raised to notify the problem.

tdunning:
OK. I will move my suggestions to a new issue.

Vincenzo-Petrolo:
@EspoTek this issue can be closed now, since #202 was opened.

PR #294, opened 2025-02-19
frequency spectrum analyzer: double sample rate support

mmehari:
This PR fixes the issue described in https://github.com/espotek-org/Labrador/issues/286.

By doubling the sample rate, the maximum frequency observed is also doubled (187.5KHz -> 375KHz).

Whereas if the **Double Sample Rate** option is not selected, the spectrum analyzer only shows half the frequencies (until 187.5KHz).

The following two pictures show the noise floor when **Double Sample Rate** is turned ON and OFF.

double_sample_rate_noise
Noise floor when **Double Sample Rate** is turned ON

single_sample_rate_noise
Noise floor when **Double Sample Rate** is turned OFF

### DIFF
--- a/Desktop_Interface/asyncdft.cpp
+++ b/Desktop_Interface/asyncdft.cpp
@@ -72,6 +72,13 @@ void AsyncDFT::addSample(short sample)
     samples_count++;
 }
 
+void AsyncDFT::clearWindow()
+{
+    window.clear();
+    data_valid = false;
+    samples_count = 0;
+}
+
 QVector<double> AsyncDFT::getPowerSpectrum(QVector<double> input)
 {
     /*Before doing anything, check if sliding DFT is computable*/
--- a/Desktop_Interface/asyncdft.cpp
+++ b/Desktop_Interface/asyncdft.cpp
@@ -75,7 +75,6 @@ void AsyncDFT::addSample(short sample)
 void AsyncDFT::clearWindow()
 {
     window.clear();
-    data_valid = false;
     samples_count = 0;
 }
 
--- a/Desktop_Interface/asyncdft.h
+++ b/Desktop_Interface/asyncdft.h
@@ -22,6 +22,9 @@ public:
     /*Add a sample to the time domain samples*/
     void addSample(short sample);
 
+    /*Clear the window*/
+    void clearWindow();
+
     /*Return the window of samples*/
     std::unique_ptr<short[]> getWindow();
 
--- a/Desktop_Interface/isodriver.cpp
+++ b/Desktop_Interface/isodriver.cpp
@@ -631,7 +631,7 @@ void isoDriver::frameActionGeneric(char CH1_mode, char CH2_mode)
 
     internalBuffer375_CH1->enableDftWrite(spectrum);
     internalBuffer375_CH2->enableDftWrite(spectrum);
-    // internalBuffer750->enableDftWrite(spectrum);
+    internalBuffer750->enableDftWrite(spectrum);
 
     internalBuffer375_CH1->enableFreqResp(freqResp, freqValue_CH1->value());
     internalBuffer375_CH2->enableFreqResp(freqResp, freqValue_CH1->value());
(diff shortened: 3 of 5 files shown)


mmehari:
@Vincenzo-Petrolo
@tdunning

tdunning:
Sweet.

There is an anomalous shift in level due to the sample rate change. This should be fixed once the vertical axis is calibrated as described in #289

mmehari:
Indeed I have see the level shift but this bug fix is all about enabling the double sample rate.

After this bug is fixed, we should tackle the calibration and unit issue and hopefully see a reduced noise floor.

tdunning:
I understand that the level calibration isn't part of this bug. The higher than desired noise floor is also out of scope here. 

My comment was simply stating that the level shift should be handled in a different issue. I think we agree on that.

mmehari:
Yes I do agree and are you able to put the PR on a test?

tdunning:
I don't have the hardware handy any more, so no, I can't help with testing. I am happy to help (a little) with math and opinions, but can't help with testing.

mi-hol:
@mmehari how do you want to proceed with this PR?

mmehari:
I was hoping either @tdunning or @Vincenzo-Petrolo would test this PR but they don't have the board. So their support is limited

I could also rely from the maintainers pool but both @EspoTek and @turboencabulator are kind of busy now.

If their won't be anyone to test this PR, then I will merge it myself.

mi-hol:
I did a quick search in closed issues and found https://github.com/espotek-org/Labrador/issues/63 
Maybe one of the posters there could be motivated to test?

tdunning:
What tests do you need?

I have the board (somewhere). I could also review screen shots if you like,
especially if you can feed a reference signal to the board.

mi-hol:
> What tests do you need?

Testing a real use case hands-on from a user perspective is required.
Looking at Screenshots has no benefit!

tdunning:
Oh well. I would have loved to have helped. I understand that some aspects are not possible to review from snaps, but I had hoped that I could add some value.

turboencabulator:
I don't really have the equipment necessary to fully test it, but I gave it a quick try with the signal generator.

One thing I noticed:  Is it storing samples in a different buffer depending on the mode?  Do we need to flush out old samples when changing the mode?

1. Set the input to, say, 1kHz.  Let it stabilize.
2. Toggle the double sample rate setting (on or off, doesn't seem to matter).
3. Change the frequency to, say, 2kHz.  Let it stabilize.
4. Toggle the double sample rate setting back.  As soon as you toggle it, the spike at 1kHz shows up temporarily, then fades.

tdunning:
That sounds like the buffer holding the 2kHz samples is reinterpreted as
having been sampled at the slower sampling rate.

As you say, it self-corrects quickly.

turboencabulator:
I don't think it's a reinterpretation because it works for any frequency pair (not just 2x like my example) and also when toggling the setting in either direction.  But it's not a big deal since it clears quickly.

mi-hol:
> Oh well. I would have loved to have helped. I understand that some aspects are not possible to review from snaps, but I had hoped that I could add some value.

Appreciate the offer, there will be a next opportunity :)

mmehari:
> One thing I noticed: Is it storing samples in a different buffer depending on the mode? Do we need to flush out old samples when changing the mode?

That is what it appears. When changing frequency values the buffers are holding both frequency samples (momentarily) and that is what is shown in the frequency spectrum.

Like you said, it clears out quickly and we can leave it as is or flush the buffer every time we change the frequency

mi-hol:
> flush the buffer every time we change the frequency

would be my preference due to the better user experience

mmehari:
Apparently there was a routine to clear oscilloscope buffers and I have extended it to clear the async_dft buffers as well. https://github.com/espotek-org/Labrador/pull/294/commits/36765ecb2aab7b7761e232f244572a713d489a31

mmehari:
@turboencabulator Did the previous two fixes solve the issue?

turboencabulator:
I think so.  There are some interesting effects near DC now when switching modes, but maybe that's expected until we read enough samples.

mmehari:
Indeed that is the case but regarding this PR I will merge it and continue further development

PR #295, opened 2025-02-21
Warning fixes

turboencabulator:
This is another batch of warning fixes:  Signed/unsigned comparisons, and some Qt deprecations.

The last commit only rearranges things to reduce how many times the compiler complains.  Qt's scroll wheel events got reworked (with a lot of the old methods getting immediately deprecated) in the 5.12 through 5.15 timeframe and we'll eventually need to address that properly.  I don't know what minimum Qt version we want to support.

### DIFF
--- a/Desktop_Interface/isodriver.cpp
+++ b/Desktop_Interface/isodriver.cpp
@@ -269,20 +269,21 @@ void isoDriver::setVoltageRange(QWheelEvent* event)
 
 void DisplayControl::setVoltageRange (QWheelEvent* event, bool isProperlyPaused, double maxWindowSize, QCustomPlot* axes)
 {
+    double steps = event->delta() / 120.0;
     if (!(event->modifiers() == Qt::ControlModifier) && event->orientation() == Qt::Orientation::Vertical) {
         double c = (topRange - botRange) / (double)400;
 
         QCPRange range = axes->yAxis->range();
 
         double pixPct = (double)100 - ((double)100 * (((double)axes->yAxis->pixelToCoord(event->y())-range.lower) / range.size()));
-        if (pixPct < 0) pixPct = 0; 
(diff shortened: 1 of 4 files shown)


mi-hol:
> minimum Qt version we want to support

I'd combined that with the question of minimum OS support.
From the [download stats ](https://github.com/espotek-org/Labrador/discussions/285#discussioncomment-12237096) Windows, Linux_AppImage and MacOS had the most downloads. 
Personally I'd draw the line for "what should we support" related to what versions of Windows,  Debian and MacOS  are still getting security patches.
Based on  https://endoflife.date I'd believe that to be Windows 10, Debian 11 and MacOS  13

Would that allow to answer your question?

mi-hol:
the situation for QT is according to https://endoflife.date/qt  a bit more complicated.

"The [KDE Project](https://kde.org/) maintains a patch collection to support [Qt 5.15 LTS](https://community.kde.org/Qt5PatchCollection) for open-source users. Moreover, [the code of commercial releases is made open within 12 months](https://kde.org/community/whatiskde/kdefreeqtfoundation/) and some level of security support still seems to continue [in the form of code patches](https://www.qt.io/blog/security-advisory-potential-integer-overflow-in-qts-http2-implementation). This does not reflect in the table above as we only track official releases from https://www.qt.io/."

turboencabulator:
The Linux CI job runs on Ubuntu 20.04 which ships with Qt 5.12.  It'll go EOL at the start of April (the CI runners get decommissioned at the same time) and we'll have to switch it to 22.04 which ships with Qt 5.15.  Maybe we should drop 5.12 support at that time?

Even though Qt 5.15 goes EOL soon after that, I imagine that most Linux distros within the past few years still offer Qt 5.15 and we should keep supporting it for a bit longer.

PR #297, opened 2025-02-26
Add download stats to Readme 

mi-hol:
@espotek-org/maintainers  Please review and comment on my proposal

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,7 @@
+[![GitHub Downloads (specific asset, specific tag)](https://img.shields.io/github/v/release/espotek-org/Labrador?style=for-the-badge)(https://img.shields.io/github/downloads/espotek-org/Labrador/continuous-win32/Labrador_Installer.exe)
+
+(https://img.shields.io/github/downloads/espotek-org/Labrador/continuous-win32/Labrador_Installer.exe)
+
 [![GitHub release (latest SemVer)](https://img.shields.io/github/v/release/espotek-org/Labrador?style=for-the-badge)](https://github.com/espotek-org/Labrador/releases/latest)
 [![GitHub All Releases](https://img.shields.io/github/downloads/espotek-org/Labrador/total?logo=github&style=for-the-badge)](https://github.com/espotek-org/Labrador/releases/latest)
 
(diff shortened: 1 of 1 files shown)


turboencabulator:
I wouldn't do this yet, at least not for the "continuous" builds.  Maybe once we have a stable release?

I've been digging into the Github API to figure out uploading assets.  Each asset file has its own download counter, and that's what you're seeing here.  The current total of 9.5k is the sum of all the assets on all the releases.  The three files don't add up to 9.5k because it's finding stats from other files from other, older releases.  I did some quick math and once you include everything, those all add up to 9.5k.

The problems:
1. Once we get the Github workflows figured out and building/uploading properly, it will delete the old file and upload a new one.  The new one starts with a download count of 0, and the old one disappears along with its download stats.  That's why the Linux AppImage is currently at 3 downloads instead of hundreds/thousands, it was recently replaced with a fresh build.  Once we get the Windows build working, we'll "lose" another 4800 downloads and look worse.  I don't see a way to carry forward the download counter, there is no way to "update" an existing file, you can only add or delete files.
2. For the AppImage in particular, the filename contains part of a git hash.  It'll change with each build, which means we would need to fix the link in the README to point to the new file on every build.  Or we could rename the file to remove that part of the git hash, but personally I like having it there as a version indicator.
3. This is only tangentially related... Github keeps a pointer to whatever release it considers "latest", and in our case it's the Android build.  Clicking the badge with the total downloads takes you there.  We should probably reorganize these releases at some point.

mi-hol:
> Maybe once we have a stable release

@turboencabulator would you perhaps have any indication by when a "stable release" could become available? 
 2 weeks ago I was under the impression  it would happen quite soon, but maybe I got this wrong.

turboencabulator:
Maybe when we get the automated builds working?  (Still missing Windows and Android, but getting close on Windows.)  What does everyone else think?

EspoTek:
I'm with Kyle on this one - not sure download stats work well with a rolling release like what we've got.

Not sure a "stable" release makes sense given that we don't have a QA team either (it's just an arbitrarily chosen release at that point).

mi-hol:
> Not sure a "stable" release makes sense given that we don't have a QA team either (it's just an arbitrarily chosen release at that point).

Just to be sure we are on the same page I've summarized what **could** be the requirements for a "stable" release:

- build all "executables" at the same time from the same version of source code
- establish a QA procedure each "executable" needs to pass before it is considered "stable"

Would above make sense? 
Anything to add or change?

@espotek-org/maintainers please comment

PR #298, opened 2025-03-05
Frequency spinbox updates

turboencabulator:
This updates the min/max frequency spinboxes on the spectrum mode to match those for the frequency response mode:  SI-prefixed units listed next to the number, and adaptive step sizes for the up/down buttons.  I've also increased the default step size for a better feel (so that the first click increments by about the same amount as the second click, instead of by 1).


mmehari:
All proposed changes are important and I don't have objection if all got merged.

turboencabulator:
There are some merge conflicts I need to resolve, and if I do it now then #299 will most likely conflict.  Once #299 is merged, I'll get this in.

PR #301, opened 2025-03-13
Fixes for Raspberry Pi

turboencabulator:
I just tested out building/running on a fresh install of Raspbian 9 (Stretch), the oldest version that we're currently claiming to support.  It needed a few tweaks to fix some compilation errors and for apparently some differences in Qt's string-to-integer conversions.

Now there should be nothing special required to build on Raspberry Pi, just the regular `qmake; make; sudo make install`.  The bootstrap script's installation of GCC 8.1 doesn't appear to be necessary since it builds fine with Raspbian's GCC 6.3 and since we're not (yet) using any C++17 features, so I removed it.


EspoTek:
Nice!  Ship it!

PR #195, opened 2021-11-13
Move data files to /usr/share on Linux

turboencabulator:
Basic idea is to make it easier to create a native package on Linux, but the program is searching for data files relative to its own location.  Instead use QStandardPaths to find the data files so we can place everything in their customary locations.

The last few commits fix some complaints by lintian for the Debian package.

PR #306, opened 2025-03-21
Filesystem Hierarchy Standard improvements

turboencabulator:
This set of changes moves installed files to their customary location on Linux:  Binaries to `/usr/bin`, data files to a subdirectory of `/usr/share`.  This goal was a large part of #195 and will make it much easier to create a native package.

Additionally `PREFIX` on the `qmake` command line now works as intended (defaulting to `/usr/local`), and the need for `APPIMAGE=1`, `DEB=1`, and `GIT_HASH_SHORT=$(git rev-parse --short HEAD)` have all been removed.

### DIFF
--- a/.github/workflows/mac.yml
+++ b/.github/workflows/mac.yml
@@ -20,8 +20,7 @@ jobs:
       - name: Build macOS app
         working-directory: Desktop_Interface
         run: |
-          GIT_HASH_SHORT=$(git rev-parse --short HEAD)
-          qmake CONFIG+=release PREFIX=/usr DEFINES+=GIT_HASH_SHORT=$GIT_HASH_SHORT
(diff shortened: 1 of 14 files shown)


turboencabulator:
I tested it on several machines and everything is working.  A few notes:
* Depending on where `libdfuprog` is installed, the run-time linker might have trouble finding it (especially now that it defaults to installing it in `/usr/local/lib`).  Seems that the linker assumes all libraries are in /usr/lib unless you tell it otherwise.  I added a `sudo ldconfig` step to the README to take care of this, should only be necessary when `PREFIX` != `/usr`.
* As always, if you're developing, it helps to have a copy installed so that it finds the libraries, firmware, waveforms, etc. in the expected places.

PR #315, opened 2025-04-01
ARM aarch64 build update + libdfuprog library

mmehari:
The current build option for ARM only targets 32-bit variants, which is inherently stated in the libdfuprog library.

`Desktop_Interface/build_linux/libdfuprog/lib/arm/libdfuprog-0.9.so: ELF 32-bit LSB shared object, ARM, EABI5 version 1 (SYSV), dynamically linked, BuildID[sha1]=bfd11ae8433823b98a0fb6fc6f3f3377ba6abc7e, with debug_info, not stripped
`

This makes 64-bit build process to fail, for example in raspberry pi 5.

In this PR, I have added a second option for compiling aarch64 architectures and included a new 64-bit libdfuprog library.


Desktop_Interface/Labrador.pro
```diff
@@ -137,17 +137,30 @@ unix:!android:!macx{
     PKGCONFIG += libusb-1.0  ##make sure you have the libusb-1.0-0-dev package!
     PKGCONFIG += fftw3       ##make sure you have the libfftw3-dev package!
     PKGCONFIG += eigen3      ##make sure you have the libeigen3-dev package!
-    contains(QT_ARCH, arm) {
-        message("Building for Raspberry Pi")
+    contains(QT_ARCH, aarch32) {
```
turboencabulator, line 140: I wish `QT_ARCH` was better documented.  On my 32-bit RPi it's exactly `arm`, and I just kicked off a job on one of Github's runners and it says `arm64`.  So I think those are the values we need to check for, `arm64` first since we're using `contains`.

Desktop_Interface/Labrador.pro
```diff
@@ -137,17 +137,30 @@ unix:!android:!macx{
...
+        message("Building for Raspberry Pi-32 bit")
         #libdfuprog include
-        LIBS += -L$$PWD/build_linux/libdfuprog/lib/arm -ldfuprog-0.9
+        LIBS += -L$$PWD/build_linux/libdfuprog/lib/arm32 -ldfuprog-0.9
```
turboencabulator, line 143: It would be convenient if the names of these directories matched `QT_ARCH`, I think we could get rid of a lot of the duplication in this file.  In addition to `arm` and `arm64`, I see `x86_64` and `i386` on a couple of my machines.

turboencabulator (changes requested):
I don't have the hardware to test it, but I think this will work once we check for the right `QT_ARCH`.

Desktop_Interface/Labrador.pro
mmehari: Indeed that makes sense and I just pushed a commit to clarify the naming ambiguity. https://github.com/espotek-org/Labrador/pull/315/commits/b6937f69e40c5bccc7769ffa2f760f820a2b1868

Desktop_Interface/Labrador.pro
mmehari: Indeed both arm architectures can be specified together. https://github.com/espotek-org/Labrador/pull/315/commits/b6937f69e40c5bccc7769ffa2f760f820a2b1868

mmehari:
I have made the change https://github.com/espotek-org/Labrador/pull/315/commits/b6937f69e40c5bccc7769ffa2f760f820a2b1868 and let me know if this is what you meant

turboencabulator (approved):
I was thinking that `contains` did a substring match, but apparently it doesn't, so ignore that part of my previous comment.

This looks good!  I'll look into setting up a job to build it along with our other releases.

PR #316, opened 2025-04-04
Remove duplications in Labrador.pro

turboencabulator:
* The files listed in `DISTFILES` seems to be copy/pasted and has several duplicates.
* We shouldn't need to add all the `INCLUDEPATH`s to `DEPENDPATH`, qmake handles that for us automatically.
* Rearrange the non-ARM `libdfuprog-0.9.so` objects so we can use the `QT_ARCH` variable to select the right one.  Also set these files as executable.


mmehari (approved):
I have looked the changes visually as well as tested it on both Linux x64 and arm64 platforms. Looks good to me

PR #317, opened 2025-04-04
Fix warnings in spectrum code

turboencabulator:
* Several different compilers complain about using the `~` operator on `bool`.  I believe it results in a bool-to-int conversion.
* There are also warnings about the `max1` and `max2` variables which are no longer used, and appear to be related to `normalizeDFT` which is also no longer used.  This removes both.


mmehari (approved):
Indeed, all the changes are relevant.

Issue #280, opened 2025-02-10
[Admin] cleanup old branches


mi-hol:
![Image](https://github.com/user-attachments/assets/47794c08-e334-4e35-b59e-1ba0cdf5a406)

1. harvest valuable changes from old branches
2. delete old branches afterwards

EspoTek:
All done!  Left Vincenzo's branch just in case there's FFT fixes in there.  :)

mi-hol:
@EspoTek WoW! That was quick! Very much appreciated!

mi-hol:
reopened to not forget about outstanding work for branch "vincenzo1207"

mi-hol:
@espotek-org/maintainers  it seems [Vincenzo-Petrolo](https://github.com/espotek-org/Labrador/commits?author=Vincenzo-Petrolo) added DFT (described in https://github.com/kfrlib/kfr?tab=readme-ov-file#fftdft) in [this](https://github.com/espotek-org/Labrador/compare/master...vincenzo1207) branch

I have unfortunaely no clue if that is beneficial for Labrador, therefore I'd like to ask one of you to step in please.
Any volunteer?

mmehari:
I don't know how you connected [Vincenzo-Petrolo](https://github.com/espotek-org/Labrador/commits?author=Vincenzo-Petrolo) with the [kfr](https://github.com/kfrlib/kfr) project, but **Vincenzo** implemented the DFT using [fftw3](https://www.fftw.org/) library which is not used in the **kfr** project.

Anyhow, I have once looked at the changes in his [branch](https://github.com/espotek-org/Labrador/tree/vincenzo1207) and all of them appear to be in master.

Nevertheless, we can reach him out for the final verdict, which I believe will be to delete the branch.

mi-hol:
> I don't know how you connected [Vincenzo-Petrolo](https://github.com/espotek-org/Labrador/commits?author=Vincenzo-Petrolo) with the [kfr](https://github.com/kfrlib/kfr) project, but **Vincenzo** implemented the DFT using [fftw3](https://www.fftw.org/) library which is not used in the **kfr** project.

It seemed to be linked to kfr because of the added dependency to "url = https://github.com/kfrlib/kfr.git" in https://github.com/espotek-org/Labrador/commit/2dc723dc9a8a7163be96f71f0d7267eb62fdc815  but my limited understanding of these technical details is the reason to aks for your help :)



> Nevertheless, we can reach him out for the final verdict

[...]

PR #318, opened 2025-04-06
AsyncDFT cleanup

turboencabulator:
Here are a few refactorings of the spectrum code.  I'm working on making the spectrum code optional (for Android, which doesn't have the libraries nor the UI set up yet to use it).

* Remove the unused `AsyncDFT::threadManager` method.  I suppose nothing in the `AsyncDFT` class is actually asynchronous anymore.
* Incoming samples were being placed in a ring buffer inside the `AsyncDFT` class.  When it's time to run an FFT, we extract the samples from the class, modify them (`isoDriver::analogConvert` and subsequent code), then pass them back to the class for the FFT.  I've moved the ring buffer to the `isoBuffer` class, alongside all the other buffers, and now `AsyncDFT` only contains the FFT code.  Additionally I've changed the ring buffer to use `std::vector<short>` instead of `std::list<double>` since it's more memory efficient.
* Change it so that all samples are placed in the ring buffer, even when spectrum mode isn't active.  This makes the program more responsive when switching between o-scope and spectrum mode since it doesn't need to wait on the ring buffer to fill.  Also check the length of the buffer independently for CH1 and CH2.
* Compute the windowing function coefficients only when changing the window.  This also fixes a bug where the windowing function would be integrated twice when CH2 is enabled.

### DIFF
--- a/Desktop_Interface/asyncdft.cpp
+++ b/Desktop_Interface/asyncdft.cpp
@@ -18,6 +18,11 @@ AsyncDFT::~AsyncDFT()
 
 QVector<double> AsyncDFT::getPowerSpectrum_dBmV(QVector<double> input, double wind_fact_sum)
 {
+    /*Before doing anything, check if sliding DFT is computable*/
+    if (input.size() < n_samples) {
+        return QVector<double>();
+    }
+
     for(int i = 0; i < n_samples; i++) {
         in_buffer[i] = input[i];
     }
--- a/Desktop_Interface/asyncdft.cpp
+++ b/Desktop_Interface/asyncdft.cpp
@@ -5,11 +5,6 @@
 
 AsyncDFT::AsyncDFT()
 {
-    /*Data is not valid until we get n_samples into the window*/
-    data_valid = false;
-    /*Samples counter*/
-    samples_count=0;
-    /*Initializing time domain window to 0s*/
     /*FFTW3 inits*/
     fftw_init_threads();
     fftw_plan_with_nthreads(omp_get_max_threads() * 2);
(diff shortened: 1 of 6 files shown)


mmehari (approved):
I have seen the commits, understood the reasoning behind it, and witnessed the performance gain. The improvements are very much welcome.

mi-hol:
> * Remove the unused `AsyncDFT::threadManager` method.  I suppose nothing in the `AsyncDFT` class is actually asynchronous anymore.

was apparently added in https://github.com/espotek-org/Labrador/compare/vincenzo1207#diff-75f83b72f071f791f2fd0d5e157481705101a540eae36f9db60cdef3589e5831R38
Would this maybe indicate the need to analyse the related branch further as I pointed out []([here](https://github.com/espotek-org/Labrador/issues/280)) ?

mi-hol:
> Would this maybe indicate the need to analyse the related branch further as I pointed out [here](https://github.com/espotek-org/Labrador/issues/280) ?

@turboencabulator would you agree with Michael's view in https://github.com/espotek-org/Labrador/issues/280#issuecomment-2781446006 ?

Desktop_Interface/asyncdft.cpp
```diff
@@ -39,11 +39,6 @@ void AsyncDFT::clearWindow()
 
 QVector<double> AsyncDFT::getPowerSpectrum_dBmV(QVector<double> input, double wind_fact_sum)
 {
-    /*Before doing anything, check if sliding DFT is computable*/
-    if (input.size() < n_samples) {
-        throw std::exception();
-    }
-
```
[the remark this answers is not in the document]
turboencabulator: I had moved that check to be the caller's responsibility.  It fixed the "catching polymorphic type ‘class std::exception’ by value" warning, and made it so an exception while processing CH2 would not affect CH1.

Checking the length inside the function is much better, so I just pushed a change to move it back.  Instead of the exception, it now returns a zero-length vector.  The plotting code will clear the display when that occurs.

Desktop_Interface/isodriver.cpp
```diff
@@ -941,11 +941,9 @@ void isoDriver::frameActionGeneric(char CH1_mode, char CH2_mode)
             auto f = m_asyncDFT->getFrequencyWindow(internalBuffer_CH1->m_samplesPerSecond);
 
             /*Creating DFT amplitudes*/
-            if (converted_dt_samples1.size() == m_asyncDFT->n_samples) {
-                auto amplitude = m_asyncDFT->getPowerSpectrum_dBmV(converted_dt_samples1, m_windowFactorsSum);
-                axes->graph(0)->setData(f, amplitude);
-            }
-            if (CH2_mode && converted_dt_samples2.size() == m_asyncDFT->n_samples) {
+            auto amplitude = m_asyncDFT->getPowerSpectrum_dBmV(converted_dt_samples1, m_windowFactorsSum);
+            axes->graph(0)->setData(f, amplitude);
```
mmehari, line 945: Now, there is a possibility that f is n_samples long and amplitude is empty (QVector<double>()). So what will happen when calling setData() here? Or if it will never happen, why do we need to check the size inside getPowerSpectrum_dBmV()?

mmehari: >The plotting code will clear the display when that occurs.

I believe you have answered my question in your previous comment.

mmehari (approved):

Desktop_Interface/isodriver.cpp
turboencabulator: That is correct.  If `f` and `amplitude` differ in length, `setData` clears the display and then uses the length of the smaller vector.

PR #323, opened 2025-04-22
Fix Android build

turboencabulator:
1. Disable everything spectrum-related, since the mobile UI isn't set up for it yet, and since we don't have the fftw/eigen dependencies set up for it either.  Accomplished mostly by wrapping those parts of the code with `#ifndef DISABLE_SPECTRUM` / `#endif`, which is all meant to be temporary until it's properly supported.
2. Add the serial encoding widgets to the mobile UI, since there is code that references the `txuart` and `serialEncodingCheck_CH1` widgets.

These two changes are enough to get the code to build.  I don't know how stable or usable it is, as compared to our current Android release.

### DIFF
--- a/Desktop_Interface/Labrador.pro
+++ b/Desktop_Interface/Labrador.pro
@@ -218,9 +218,9 @@ unix:SOURCES += unixusbdriver.cpp
 unix:HEADERS += unixusbdriver.h
 
 # For multithreading on Unix fftw
-unix:!macx: LIBS += -fopenmp
+unix:!android:!macx: LIBS += -fopenmp
 macx: LIBS += -lomp
-unix: LIBS += -lfftw3_omp
+unix:!android: LIBS += -lfftw3_omp
 
 #############################################################
 ########       SHARED ANDROID/LINUX GCC FLAGS      #########
@@ -282,6 +282,11 @@ android {
     #liblog include
     LIBS += -L$$PWD/build_android/liblog/lib -llog
     ANDROID_EXTRA_LIBS += $${PWD}/build_android/liblog/lib/liblog.so
+
+    # Frequency spectrum/response disabled for now, needs UI and supporting libraries
+    DEFINES += DISABLE_SPECTRUM
+    SOURCES -= asyncdft.cpp
+    HEADERS -= asyncdft.h
 
     # Library dependencies are only compiled for this ABI currently
     ANDROID_ABIS = armeabi-v7a
--- a/Desktop_Interface/isobuffer.cpp
+++ b/Desktop_Interface/isobuffer.cpp
@@ -29,19 +29,27 @@ constexpr auto kTopMultimeter = 2048;
 constexpr double kTriggerSensitivityMultiplier = 4;
 }
 
+#ifndef DISABLE_SPECTRUM
 isoBuffer::isoBuffer(QWidget* parent, int bufferLen, int windowLen, isoDriver* caller, unsigned char channel_value)
+#else
+isoBuffer::isoBuffer(QWidget* parent, int bufferLen, isoDriver* caller, unsigned char channel_value)
+#endif
     : QWidget(parent)
     , m_channel(channel_value)
     , m_bufferPtr(std::make_unique<short[]>(bufferLen*2))
     , m_bufferLen(bufferLen)
+#ifndef DISABLE_SPECTRUM
     , m_window_capacity(windowLen)
+#endif
     , m_samplesPerSecond(bufferLen/21.0/375*VALID_DATA_PER_375)
     , m_sampleRate_bit(bufferLen/21.0/375*VALID_DATA_PER_375*8)
     , m_virtualParent(caller)
 {
     m_buffer = m_bufferPtr.get();
+#ifndef DISABLE_SPECTRUM
     m_window.reserve(m_window_capacity);
     m_window_iter = m_window.begin();
+#endif
 }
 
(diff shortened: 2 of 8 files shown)


EspoTek:
LGTM

turboencabulator:
I tried it on my phone (using some variation of `adb install [options] labrador-debug.apk`).

* When the app starts, there are 3-4 prompts to let it access USB.  This should be cleaned up, but is unrelated to this change.
* My phone doesn't have enough screen real estate to provide a good user interface.  Maybe a tablet would be better?  There's just not enough room to properly show the new text edit area and the new checkbox, and parts of them get cut off.
* I can confirm that serial decoding works.  But for encoding, the app crashes on the first character I type into the encoding box.  I don't know yet how to debug it.
* The app crashes when I rotate from portrait to landscape orientation.  Should be unrelated to this change though.

turboencabulator:
I'll go ahead and merge this to give us a starting point for the issues noted above.

mi-hol:
> I tried it on my phone (using some variation of `adb install [options] labrador-debug.apk`).

@turboencabulator I tried to install from "Artifacts" on [this completed "Action"](https://github.com/espotek-org/Labrador/actions/runs/14621545271) on a 10 years old Tablet running Android 4.4.4(KitKat) but on this old OS GH UI even didn't enabled the option to  download apk. :( 
The old [1.1 Release link ](https://github.com/espotek-org/Labrador/releases/download/android/Labrador-Signed.apk) did work though, strange.
Is there maybe another URL to download your non-release apk directly?

mi-hol:
@turboencabulator difference I see now is downloadable apk is signed (according to it's name)

turboencabulator:
> Is there maybe another URL to download your non-release apk directly?

Unfortunately the upload-artifact action [always zips the files](https://github.com/actions/upload-artifact#zip-archives), even if it is only a single file.  You'll have to download the .zip and extract the .apk from it.

> @turboencabulator difference I see now is downloadable apk is signed (according to it's name)

Signing the .apk hasn't been set up yet, I need to figure out how to do it.  Once that is ready, I'll add it to the assets on the continuous release.  Also there might be other challenges to installing something that is unsigned.

EspoTek:
> Signing the .apk hasn't been set up yet, I need to figure out how to do it. Once that is ready, I'll add it to the assets on the continuous release. Also there might be other challenges to installing something that is unsigned.

I've got a digital signature with Google if you need me to add it as an environment variable.

mi-hol:
> digital signature with Google if you need me to add it as an environment variable.

@EspoTek That sounds like the solution

mi-hol:
> You'll have to download the .zip and extract the .apk from it.

Current Windows Defender reports the zip as containing a security thread and removes it :(

image

mi-hol:
Good news, new Android build doesn't trigger Windows Defender :)
So I was able to download apk

mi-hol:
But no luck with installing the new version on Android Table.

Opening the apk file gives the following message (translated from German)

Parsing-Error
There was a problem parsing the package

turboencabulator:
> I've got a digital signature with Google if you need me to add it as an environment variable.

@EspoTek I think I've got everything set up to sign.  We'll need the certificate and private key added to `ANDROID_CERTIFICATE_BASE64` and `ANDROID_PRIVATE_KEY_BASE64` respectively.  I left some comments in the android workflow for how to extract them out of a .keystore or equivalent file, assuming that's how you have them stored.

mi-hol:
@turboencabulator did you see my suggestion for consistent asset naming [here](https://github.com/espotek-org/Labrador/discussions/282#discussioncomment-12875885)?

turboencabulator:
> @turboencabulator did you see my suggestion for consistent asset naming [here](https://github.com/espotek-org/Labrador/discussions/282#discussioncomment-12875885)?

Yes, I agree and it's on my to-do list.  I just want to make sure renaming things doesn't break anything.

EspoTek:
> > I've got a digital signature with Google if you need me to add it as an environment variable.
> 
> @EspoTek I think I've got everything set up to sign. We'll need the certificate and private key added to `ANDROID_CERTIFICATE_BASE64` and `ANDROID_PRIVATE_KEY_BASE64` respectively. I left some comments in the android workflow for how to extract them out of a .keystore or equivalent file, assuming that's how you have them stored.

Finally catching up on old messages, have added both of these variables to the Repo secrets.

turboencabulator:
Thanks @EspoTek, but no luck getting it to sign yet.  I think there is something wrong with the format of the secret and it's difficult to debug while not exposing the secret...

Whatever values make these commands happy should be the right format:
```
# These first two should print the contents in human-readable format
# !!! Don't put these commands into any workflow !!!
echo "$ANDROID_CERTIFICATE_BASE64" | base64 --decode | openssl x509 -inform DER -noout -text
echo "$ANDROID_PRIVATE_KEY_BASE64" | base64 --decode | openssl pkey -inform DER -noout -text

# These next two should create two identical files when the key is in PKCS#8 format:
echo "$ANDROID_PRIVATE_KEY_BASE64" | base64 --decode > a.p8
echo "$ANDROID_PRIVATE_KEY_BASE64" | base64 --decode | openssl pkcs8 -inform DER -outform DER -topk8 -nocrypt > b.p8
diff a.p8 b.p8  # Should be identical
```

EspoTek:
Does it work now?

turboencabulator:
> Does it work now?

Yes it does!  I'll get the Android build added to the continuous release.

PR #332, opened 2025-05-04
Build system cleanup

turboencabulator:
The main goal is to eliminate a bunch of manual steps in the build process and let `qmake` do most of the work for us.  I've kicked off a few builds for testing these changes, please try the four builds against the `build-cleanup` branch at https://github.com/espotek-org/Labrador/actions

All platforms:
* All assets in the release now have names of the form "Labrador-version[-type].ext", where "version" is the Git short hash, and the optional "type"  is any extra info to disambiguate anything that isn't clear from the file extension (e.g. CPU architecture, that the .exe is an installer and not a standalone program, etc.).
* Move firmware, waveforms, and images to the `resources` directory in the repo.  The copies in the `build_android` directory were removed so they won't get out-of-sync.
* Move icons and other OS-specific files to the relevant `build_*` directory.
* Built targets (particularly the Labrador executable) are no longer placed in the `bin` subdirectory.

Linux:
* The executable is now lowercase `labrador` to match existing uses (in documentation, in scripts, in .desktop files, etc.), and to follow the usual conventions for that platform.
* Move the `debian` directory up one level, to where the Debian tools should expect to find it.

macOS:
* The `make` step now takes care of installing the firmware, waveforms, icon, and Info.plist (which is now automatically generated) into the `Labrador.app` directory.
* Firmware and waveforms moved from `Contents/MacOS` to `Contents/Resources`, where the QStandardPaths module should now look for them.  Needs testing since I don't have a mac.
* There are a few new items appearing in the generated Info.plist (notably the `QMAKE_MACOSX_DEPLOYMENT_TARGET = 10.10`), and I noticed that the reference to the icon now has the ".icns" suffix where prevously it didn't.  Also needs testing to ensure this is all correct.
* The [volume name](https://doc.qt.io/qt-5/macos-deployment.html#volume-name) should no longer have the "bin/" prefix.
* No longer bundles the `dfu-programmer` Windows executable.
* We were linking to our copy of libusb, but `macdeployqt` would bundle the system's libusb.  I've removed our copy, we'll use the system's copy for both to avoid mixing and matching.

Windows:
* New `nmake INSTALL_ROOT=wherever install` step now handles copying the DLLs, firmware, and waveforms to the desired location.  Previously it was relying on the `nmake` command to place the built target in the `bin` directory alongside the firmware/waveforms.
* Now using the QStandardPaths module to find the firmware and waveforms, which should look in the same directory as the executable.


Desktop_Interface/Labrador.pro
```diff
@@ -101,22 +98,38 @@ win32 {
...
+
+    waveforms.path = /waveforms
+    waveforms.files = $$files(resources/waveforms/*)
+
+    INSTALLS += target lib_deploy firmware waveforms
+
     #libusbk include
     contains(QT_ARCH, i386) {
-        CONFIG(release, debug|release): LIBS += -L$$PWD/build_win/libusbk/bin/lib/x86/ -llibusbK
-        else:CONFIG(debug, debug|release): LIBS += -L$$PWD/build_win/libusbk/bin/lib/x86/ -llibusbK
```
turboencabulator, line 107: These two lines both point to the same library, and I realized later that maybe it was intended for one to link to the `.dll`, and the other to the `.lib`?

turboencabulator:
> the external artifact name currenctly always starts with "asset", would it be possible to replace that with "labrador" too?

I'd prefer not, at least with how the download-artifact action lets us [pick a subset of artifacts](https://github.com/actions/download-artifact#download-multiple-filtered-artifacts-to-the-same-directory).  Everything starting with "asset" gets added to the release assets, and changing it to "labrador" would be too generic since it doesn't identify the type of the artifact.

EspoTek:
This sounds cool, cheers man!

PR #335, opened 2025-06-03
Improvements to cursor measurement display

turboencabulator:
The string containing the cursor measurements tends to be too long to fit inside the plot window, and the left side gets cut off.  I've added some line breaks, right-alignment, and made the font a bit smaller so that now the measurements fit nicely in the top right corner.  Plus some other code cleanup.
Screenshot at 2025-06-02 23-15-36

### DIFF
--- a/Desktop_Interface/isodriver.cpp
+++ b/Desktop_Interface/isodriver.cpp
@@ -636,8 +636,8 @@ void isoDriver::updateCursors(){
 #endif
     if (!cursorStatsEnabled) return;
 
-    char temp_hori[64];
-    char temp_vert[64];
+    QString temp_hori;
+    QString temp_vert;
 #ifndef DISABLE_SPECTRUM
     if(spectrum)
     {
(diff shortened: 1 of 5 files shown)


EspoTek:
Ship it.

mmehari:
I like the improvements and I have one small comment.

Depending on the number of digits of the measurement values, the display goes left and right when moving the cursor.
Is it possible to fix the alignment? For example, the colon symbol (:) can be used as an alignment marker.

turboencabulator:
The alignment might be touching on a bigger discussion.  While testing, I noticed it will show a value like `999.99mV` (5 digits of precision in 8 characters) and after incrementing slightly, it will show `1.00V` (3 digits of precision in 5 characters).  I'd like to change it to a constant 3 or 4 digits of precision, using format strings `%.3g` or `%.4g` respectively in `siprint.cpp`.  Any preferences?  I like the way 3 digits look, but I also like having a 4th digit.

mmehari:
4 digit precision is fine by me as long as the alignment is fixed

turboencabulator:
I've aligned all lines at the colon.  Give it a try now.

mmehari:
It is perfect.

PR #342, opened 2025-07-27
disable creating of issues without the predefined & required content

mi-hol:
to avoid missing information about OS & version in issues


EspoTek:
I completely agree that disabling blank issues is a good idea, but before we do this I think we might want to edit the Bug Report template to request OS and version from the user.

mi-hol:
> I completely agree that disabling blank issues is a good idea, but before we do this I think we might want to edit the Bug Report template to request OS and version from the user.

good catch. Added it now

mi-hol:
@EspoTek any objections to merge latest version?

Issue #343, opened 2025-07-27
Labrador app not installable on supported Android 64bit versions


mi-hol:
> > I know, I've tried Labrador-82ea67c-release-signed.apk on my Pixel 9 Pro/Android 15. "App not installed as app isn't compatible with your phone."
> 
> It looks like this is probably a 32-bit vs 64-bit issue. We'll probably need to get both versions going.
> 
>  @EspoTek @turboencabulator could one of you please elaborate on activities to likely fix this issue? 

 _Originally posted by @mi-hol in [#249](https://github.com/espotek-org/Labrador/issues/249#issuecomment-3124583172)_

mi-hol:
apk file currently contains only "\lib\armeabi-v7a" 32bit libraries.
64bit devices seem to require "arm64-v8a" libraries.

mi-hol:
[/Desktop_Interface/Labrador.pro](https://github.com/espotek-org/Labrador/blob/master/Desktop_Interface/Labrador.pro#L324) contains the reference:

```
# Library dependencies are only compiled for this ABI currently
ANDROID_ABIS = armeabi-v7a

```

@espotek-org/maintainers would this be the place to enhance for enabling 64bit support?

mi-hol:
found explanation on https://developer.android.com/ndk/guides/abis#arm64-v8a

mi-hol:
I've created branch [android_abi](https://github.com/espotek-org/Labrador/blob/android_abi/Desktop_Interface/Labrador.pro#L325) and tried to compile for arm64-v8a but that fails with error:

`/usr/local/lib/android/sdk/ndk/21.3.6528147/toolchains/arm-linux-androideabi-4.9/prebuilt/linux-x86_64/lib/gcc/arm-linux-androideabi/4.9.x/../../../../arm-linux-androideabi/bin/ld: error: cannot find -lusb1.0`

Would perhaps somebody know how to fix it?

EspoTek:
> I've created branch [android_abi](https://github.com/espotek-org/Labrador/blob/android_abi/Desktop_Interface/Labrador.pro#L325) and tried to compile for arm64-v8a but that fails with error:
> 
> `/usr/local/lib/android/sdk/ndk/21.3.6528147/toolchains/arm-linux-androideabi-4.9/prebuilt/linux-x86_64/lib/gcc/arm-linux-androideabi/4.9.x/../../../../arm-linux-androideabi/bin/ld: error: cannot find -lusb1.0`
> 
[...]

PR #350, opened 2025-08-26
Update Windows workflow to use Windows 2022

EmilZackrisson:
Update Windows workflow from the unsupported windows-2019 to the new windows-2022.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -6,7 +6,7 @@ on:
 
 jobs:
   build-windows:
-    runs-on: windows-2019
+    runs-on: windows-2022
     strategy:
       fail-fast: false
       matrix:
@@ -79,7 +79,7 @@ jobs:
   build-installer:
     needs:
       - build-windows
-    runs-on: windows-2019
+    runs-on: windows-2022
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -39,6 +39,7 @@ jobs:
         uses: ilammy/msvc-dev-cmd@v1
         with:
           arch: ${{ matrix.platform }}
+          toolset: 14.2
 
       - name: Build librador
         if: matrix.platform == 'x64'
--- a/.gitignore
+++ b/.gitignore
@@ -51,3 +51,6 @@ Librador_API/___librador/librademo/librademo
 
 Librador_API/___librador/basicdemo/Makefile*
 Librador_API/___librador/basicdemo/basicdemo
+
+Labrador.app
+.DS_Store


mi-hol:
@EmilZackrisson thanks for your contribution!

I did not see an invocation of this action in your fork.
Has the changed action been running successfully?

EmilZackrisson:
@mi-hol 
> I did not see an invocation of this action in your fork. Has the changed action been running successfully?

Yes, they ran successfully. I have rerun them so you can see.

mi-hol:
@espotek-org/maintainers I downloaded the build Windows 64bin.zip and opened it successfully but didn't perform functional tests.
Please let me know any objections within the next 6 days.
I'd plan to merge it after this timeline expiered.

EspoTek:
No way, was this really all we needed to do all along?

@mi-hol let me test it on a clean install of Windows but any potential issue caused by this change should cause a crash at boot - if it launches at all we should be good.

EspoTek:
I ran the file through ChatGPT to try and work out why it didn't explode into flames, apparently there's binary compatibility between VS2019 and VS2022 redistributables:
https://chatgpt.com/share/68ae9b73-4578-8007-83c3-7eeadda2a203

However, to be 100% kosher, it recommends we force the toolset to v142 (so that it will build in "VS2019 compatibility mode"):
```
- uses: ilammy/msvc-dev-cmd@v1
  with:
    arch: ${{ matrix.platform }}
    toolset: 14.2   # selects the latest v142 toolset installed
```

We had a whole discussion about how to tackle this perceived incompatibility on the maintainers board a couple of months back, it looked like it was going to be a nightmare.

Thanks for the solution Emil, simple is best!

mi-hol:
> We had a whole discussion about how to tackle this perceived incompatibility on the maintainers board a couple of months back, it looked like it was going to be a nightmare.

@EspoTek unfortunately I have no recollection regarding a "nightmare" and the only reference I found in discussions is https://github.com/espotek-org/Labrador/discussions/282#discussioncomment-12346980.
Maybe I'm missing something important and you could kindly help me out with a link?

EspoTek:
> > We had a whole discussion about how to tackle this perceived incompatibility on the maintainers board a couple of months back, it looked like it was going to be a nightmare.
> 
> @EspoTek unfortunately I have no recollection regarding a "nightmare" and the only reference I found in discussions is [#282 (reply in thread)](https://github.com/espotek-org/Labrador/discussions/282#discussioncomment-12346980). Maybe I'm missing something important and you could kindly help me out with a link?

I just checked the discussions; both Kyle and I were under the impression we'd need to completely move away from MSVC to get it building with Qt5, or manually set up VS2019 on the 2022 server.  I actually spent a couple of hours trying to solve the latter.

But nope, the correct solution was literally just changing the GitHub runner from 2019 to 2022 and clicking build!

mi-hol:
> both Kyle and I were under the impression we'd need to completely move away from MSVC to get it building with Qt5, or manually set up VS2019 on the 2022 server.

@EspoTek The way I read https://github.com/espotek-org/Labrador/discussions/282#discussioncomment-12346749 it resulted in the _unconfirmed_ assumption that current source code has a dependency on VS2017/VS2019, right?

mi-hol:
> However, to be 100% kosher, it recommends we force the toolset to v142 (so that it will build in "VS2019 compatibility mode"):

From my view this would add a **technical debt** (similar to the specific USB lib version for Android https://github.com/espotek-org/Labrador/issues/343#issuecomment-3194623238), that at this stage (without bug reports) seems unwarranted.
I stand to be corrected but would like to see some evidence for taking that route.

mi-hol:
@espotek-org/maintainers I plan to merge this PR tomorrow 5Sep2025 unless concerns are logged here.

EspoTek:
> > However, to be 100% kosher, it recommends we force the toolset to v142 (so that it will build in "VS2019 compatibility mode"):
> 
> From my view this would add a **technical debt** (similar to the specific USB lib version for Android [#343 (comment)](https://github.com/espotek-org/Labrador/issues/343#issuecomment-3194623238)), that at this stage (without bug reports) seems unwarranted. I stand to be corrected but would like to see some evidence for taking that route.

Nah, if anything it's the opposite.  Qt5 using Microsoft compilers requires VS2019 (or earlier) runtime, so if we moved to a platform that didn't allow us to build with toolset v142 then that line would flag the error immediately.  I'll add it to the PR now, then test and merge.

EspoTek:
Yep, it works great.  Thanks for your contribution, Emil!

mi-hol:
> Qt5 using Microsoft compilers requires VS2019 (or earlier) runtime

I checked the official QT 5.15 LTS doc [here ](https://doc.qt.io/archives/qt-5.15/windows.html#supported-configurations) and it says:


Operating System | Architecture | Compiler | Notes
-- | -- | -- | --
Windows 11 22H2 | x86 and x86_64 | **MSVC 2022**, MSVC 2019, MSVC 2017, MSVC 2015, MinGW 8.1


Hence I failed to see above quoted requirement, would you mind to enlighten me where this is stated please? This will allow me to document the known technical debts with proper references.

EspoTek:
>  we can easily move to VS2022 if it's required in future.

With the common IT obsolesence cycles, as former IT product manager, I learned that the future is now.
In other words: It doesn't pay out to postphone technical infrastructure changes because once done the costs to fix them quickly accumulate to an unsurmountable amount

Issue #287, opened 2025-02-13
It should be possible to zoom the spectrum analyzer in frequency and amplitude


tdunning:
Currently, the vertical and frequency axes in the spectrum are what you get. It would be much better if you could zoom in on a particular range of amplitudes or on a particular range of frequencies.

The data should be available so this should be doable with just a UI action.

I

PR #352, opened 2025-09-05
Zoom feature for the spectrum analyzer and frequency response tools

brentfpage:
Implemented #287  .  Also added the possibility of log-spacing the frequency axis.  The ticking functionality for the log-spaced frequency axis is decent, but I'm open to suggestions for improvement.  Notably, it seems that the available QCPAxis ticking functions are described by https://web.archive.org/web/20160904175553/http://www.qcustomplot.com/documentation/classQCPAxis.html instead of the most recent version of that documentation.

I've only tested the new code on Mac .  Judging by DisplayControl::setVoltageRange and MainWindow::on_actionShow_Range_Dialog_on_Main_Page_triggered, in order to get the new zoom features to work on Android, it may be necessary to add a few functions like topRangeUpdated but for the horizontal direction.

Because the zoom level desired by a user of the Frequency Response tool may not match the min. and max. frequency of the desired frequency response, I kept the min. and max. frequency selectors for this tool while also enabling fancy zoom with the cursor.

### DIFF
--- a/Desktop_Interface/isodriver.cpp
+++ b/Desktop_Interface/isodriver.cpp
@@ -1069,7 +1069,6 @@ void isoDriver::frameActionGeneric(char CH1_mode, char CH2_mode)
         axes->yAxis->setLabel("Relative Power (dBmV)");
         axes->xAxis->setRange(display->leftRange, display->rightRange);
         axes->yAxis->setRange(display->botRange, display->topRange);
-        /*Setting maximum/minimum y-axis -60dBmV to 90dBmV*/
 
     } else if (freqResp) {
         if (!paused_CH1) {
--- a/Desktop_Interface/isodriver.h
+++ b/Desktop_Interface/isodriver.h
@@ -215,7 +215,6 @@ private:
     //Spectrum
     int m_spectrumCounter = 0;
     AsyncDFT *m_asyncDFT;
-    bool m_specrumLogF = false;
     double m_spectrumMinY = -60;
     double m_spectrumMaxY = 90;
     int m_windowingType = 0;


mmehari:
Very nice. We now have the zoom feature to the remaining oscilloscope plots.

I have tested it on Ubuntu 24.04 and it is working fine.

One question though: what is **freq. spacing**? I sometimes see it on the bottom right corner and I wonder what it is used for

brentfpage:
Thanks!  The freq. spacing label just indicates the multiplicative difference between adjacent x-axis ticks when the x-axis is log-spaced.  I figured it was useful to display as an extra reference point when the plot is zoomed in to the extent that the x ticks aren't simple numbers like A x 10^n, but maybe it's overkill.

mmehari:
Now I understand the **freq. spacing** label and I don't think it is an overkill feature.

I am OK if the zoom feature got merged to the project.

brentfpage:
Am I able to do that?  Think I probably don't have the right permissions.

mmehari:
Let us give a chance for others to comment on it. And if no one responds until Wednesday, I will merge it myself.
@espotek-org/maintainers

brentfpage:
Ah ok got it

EspoTek:
> There are lots of commented code lines that don't make sense to me. Either they are intended for debugging and require a matching comment to make sense or are to be removed.

It looks like test code to me.  I agree that ideally we should get rid of it before merge but in general the commit is nice!

brentfpage:
> It looks like test code to me.

Yes that and some lines that were becoming obsolete that I wasn't comfortable deleting quite yet.
Also, referring to my original comment, I was misinterpreting the purpose of DisplayControl::setVoltageRange and MainWindow::on_actionShow_Range_Dialog_on_Main_Page_triggered.  I now have no reason to believe that the new frequency response and spectrum zoom features won't work on Android.
> but in general the commit is nice!

Thank you!

EspoTek:
> I see the file's (Tinylab_proto1.kicad_pcb) change in the commit message but it doesn't appear in the Files changed section. I don't know how that happens.

It looks like it was changed then reverted (see the commit history).  I'm going to merge this for now!

Thanks Brent!

brentfpage:
My pleasure!  Sorry about that temporarily edited file in the fork (Tinylab_proto1.kicad_pcb).  I committed with it edited, realized my mistake, then reverted it to the original version.

mi-hol:
@brentfpage build of Android fails now. Could this be related to the changes?

```
void MainWindow::paused(bool enabled)
                             ^
isodriver.cpp:413:21: error: no member named 'retickXAxis' in 'isoDriver'
            driver->retickXAxis();
```

brentfpage:
'retickXAxis' is a function that I added and looks to be responsible given that error message.  It's declared in isodriver.h within a "#ifndef DISABLE_SPECTRUM" block , so maybe lines 412-414 of isodriver.cpp should also be wrapped in such a block?  This issue would never arise at runtime because all calls to DisplayControl:setRespAndSpecRanges (which includes lines 412-414 of isodriver.cpp) are already wrapped in "#ifndef DISABLE_SPECTRUM" statements, but maybe the compiler is not aware of that fact and so is complaining.

brentfpage:
I was able to reproduce the error when compiling on mac by adding DISABLE_SPECTRUM to the defines.  Wrapping the declaration of DisplayControl:setRespAndSpecRanges (line 48 of isodriver.h) as well as the definition of this function (lines 351-418 of isodriver.cpp) in "#ifndef DISABLE_SPECTRUM" ... "#endif" guards then fixed the error.  I can add a new PR with these changes if that would be convenient.

mi-hol:
> add a new PR with these changes

is appreciated

brentfpage:
Done in #356

Issue #354, opened 2025-09-18
Merge requests don't build all assets automatically, issues created go unnoticed


mi-hol:
### PROBLEM DESCRIPTION

Merge requests don't build all assets automatically, issues created go unnoticed

### REQUESTED INFORMATION

- [x] Verified with latest published [Labrador version](https://github.com/espotek-org/Labrador/releases)
- [x] Searched the problem in the [Troubleshooting Guide](https://github.com/espotek-org/Labrador/wiki/Troubleshooting-Guide)
- [ ] Searched the problem in the [docs](https://github.com/espotek-org/Labrador/wiki)
- [ ] Searched the problem in [issues](https://github.com/espotek-org/Labrador/issues)
- [ ] Searched the problem in [discussions](https://github.com/espotek-org/Labrador/discussions)

- Hardware the software is running on: (i.e. copied from output of Windows command *systeminfo* "System Type: x64-based PC" )
- Operating System (OS) Build Version: (i.e. "OS Name: "Microsoft Windows 11 Pro" & "OS Version: 10.0.22621" & "OS Configuration:      Standalone Workstation")
- Labrador Git hash:  (from menu "Help/About Espotek Labrador)

### TO REPRODUCE
_Steps to reproduce the behavior:_

1. Go to 'GitHub/Actions'
2. Click on 'Build Android'
3. Select 'Run workflow'
4. See error i.e. https://github.com/espotek-org/Labrador/actions/runs/17824122006

### EXPECTED BEHAVIOUR
A PR should trigger all build workflows

mi-hol:
fixed via https://github.com/espotek-org/Labrador/pull/380

PR #357, opened 2025-09-19
fixed minor time ticking bug

brentfpage:
In some cases, switching from log-spaced frequency axis mode back to the standard oscilloscope view results in unreadable time axis ticks.  This PR fixes this bug.


mmehari:
Good catch

EspoTek:
Also good catch.  Just as a dumb thought, would setting the number format to "gb" cause weirdness on users with different locals?  I'm thinking for example Germans who use , instead of .

brentfpage:
When I query ```xAxis->numberFormat()``` and ```xAxis->numberPrecision()``` at the program initiation stage, 'gb' and 6 are returned, which is why I went with those.  Also, in ```MainWindow:initializePlot```,     ```ui->scopeAxes->xAxis->setAutoTickCount(9);``` gets called, which is why I revert ```AutoTickCount``` to 9 when switching back from the log-spaced frequency spacing.  Now I notice as well though that ```ui->scopeAxes->xAxis->setAutoTickCount(9);``` is only called in ```MainWindow:initializePlot``` when ```#if QCP_VER == 1``` is satisfied. When ```#if QCP_VER == 2``` instead, some different ticking functions are used... should I probably look into those and make sure #352 is compatible with them?

mi-hol:
> Now I notice as well though that `ui->scopeAxes->xAxis->setAutoTickCount(9);` is only called in `MainWindow:initializePlot` when `#if QCP_VER == 1` is satisfied. When `#if QCP_VER == 2` instead, some different ticking functions are used... should I probably look into those and make sure #352 is compatible with them?

@espotek-org/maintainers could one of you please comment?

EspoTek:
> > Now I notice as well though that `ui->scopeAxes->xAxis->setAutoTickCount(9);` is only called in `MainWindow:initializePlot` when `#if QCP_VER == 1` is satisfied. When `#if QCP_VER == 2` instead, some different ticking functions are used... should I probably look into those and make sure #352 is compatible with them?
> 
> @espotek-org/maintainers could one of you please comment?

I don't think this is a big problem, there were plans to use QCustomPlot v2 but this was adandoned.

Addressing the original comment though, I think it would be safer to query `xAxis->numberFormat()` during initialisation, save it to a member variable and reset it later, rather than using a hardcoded value of "gb".

brentfpage:
@EspoTek Is the new commit what you had in mind?  Some of my code in ```isodriver:retickXAxis()``` still uses hard-coded ticking settings,  but it seems acceptable in this fairly special-purpose function.
I tested the new commit and it seems to work well.

mi-hol:
@EspoTek thanks for reviewing and merging!
As there is no automatic build after merge configured yet (logged in #354).
Shall I trigger "continous build"?

EspoTek:
> @EspoTek thanks for reviewing and merging! As there is no automatic build after merge configured yet (logged in #354). Shall I trigger "continous build"?

Yes please!  Let's get this fix through ASAP.

mi-hol:
@EspoTek @brentfpage unfortunately the Mac build fails at https://github.com/espotek-org/Labrador/actions/runs/18072603968/job/51424583461#step:3:553

Maybe due to usage of unsupported MacOS version 13.
Any ideas how to resolve that are greatly appreciated.

brentfpage:
Appears likely related to the item "Drop macOS Ventura CI in Homebrew/core" in the changelog for homebrew 4.6.11, https://github.com/Homebrew/brew/releases/tag/4.6.11    .  I'm able to compile the desktop app without any issues on my local system, running mac os 15.5. So, maybe the fix could be as simple as updating ```macos-13```->```macos-15``` in the top of the ```mac.yml``` workflow file?  I don't have a good idea of how dangerous such a patch would be though.  In the Labrador.pro file in Desktop_Interface/, there's also a statement ```QMAKE_MACOSX_DEPLOYMENT_TARGET = 10.10```, but this seemingly outdated line doesn't have any effect in practice in my experience.

mi-hol:
> updating `macos-13`->`macos-15` in the top of the `mac.yml` workflow file

@brentfpage thanks for researching! Lets try that. 
Could you please run the modified action in your fork on GH as a confirmation?

brentfpage:
@mi-hol Did so and the macOS build went through successfully!  Btw, I don't think the problem is related to any specific part of the code from this PR.  Now that homebrew has made macOS 13 deprecated, use of ```macos-13``` in the ```mac.yml``` probably would cause any macOS build to fail.

mi-hol:
> Did so and the macOS build went through successfully!

Thank you so much! I'll copy your chage. 

> I don't think the problem is related to any specific part of the code from this PR. Now that homebrew has made macOS 13 deprecated, use of `macos-13` in the `mac.yml` probably would cause any macOS build to fail.

That is aligned with my initial thinking when I wrote "Maybe due to usage of unsupported MacOS version 13."

PR #360, opened 2025-09-30
Update macos to version 15 becasue 13 is now unsupported

mi-hol:
for details see https://github.com/espotek-org/Labrador/pull/357#issuecomment-3344215671

### DIFF
--- a/.github/workflows/mac.yml
+++ b/.github/workflows/mac.yml
@@ -6,7 +6,7 @@ on:
 
 jobs:
   build-dmg:
-    runs-on: macos-13
+    runs-on: macos-15
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4


brentfpage:
Argh, I should have tested this before, but when I actually download the .dmg created by the build and try to run the application it contains, I'm not able to.  A pop-up comes up that says "You can’t open the application “Labrador” because this application is not supported on this Mac."  I tried using ```runs-on: macos-14``` in the mac.yml file instead, but the application produced by the build action still did not work on my mac.
I'm surprised, because when I run these few lines from the .yml
```
brew install fftw libomp qt@5 libusb eigen
brew link libomp qt@5 --force
qmake -config release
make -j$(sysctl -n hw.ncpu)
macdeployqt Labrador.app -verbose=2 -dmg -libpath=build_mac/libdfuprog/lib/
```
on my local system running macOS 15.5, the resulting .dmg works just fine.

PR #349, opened 2025-08-26
Add baud rate for 31250

EmilZackrisson:
Added a baud rate of 31250, used in MIDI. Tested on macOS. 
Resolves #269

PR #365, opened 2025-10-12
Create osv-scanner-scheduled.yml

mi-hol:
enable scanning for security vulnerability issues

### DIFF
--- a/.github/workflows/mac.yml
+++ b/.github/workflows/mac.yml
@@ -6,7 +6,7 @@ on:
 
 jobs:
   build-dmg:
-    runs-on: macos-15
+    runs-on: macos-15-intel
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
@@ -15,7 +15,7 @@ jobs:
         run: |
           brew update
           brew remove --force $(brew list --formula) || true
-          brew install fftw libomp qt@5 libusb eigen
+          brew install fftw libomp qt@5 libusb eigen create-dmg
           brew link libomp qt@5 --force
 
       - name: Build librador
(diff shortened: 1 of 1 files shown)


mi-hol:
@EspoTek there is an open error I noticed too late regarding branch name:main.
I've fix it already in another branch that I plan to merge later.
Please merge this PR in case this is the only review issue found.

EspoTek:
I'm wondering how necessary this is, since the app is:
 - Not hugely popular
 - Doesn't communicates with the open web
 - Doesn't have many dependencies
 - Isn't highly specialised

Potentially will show some scary messages but probability is basically zero that the app could ever be used as an attack vector.

That said this is a cool tool and I'll definitely include in some of my other apps that are exposed to the open web!

Thoughts @espotek-org/maintainers ?

mi-hol:
> Potentially will show some scary messages but probability is basically zero that the app could ever be used as an attack vector.

In my test just 3 messages due to "calls to pure virtual member functions in constructors and destructors" surfaced. For details see [here](https://github.com/mi-hol/Labrador/security/code-scanning/13).
I'll fix them as the next step.

Main purpose from my view is to safeguard the project from contributions that introduce undesireable changes that could go otherwise unnoticed (example is in https://github.com/espotek-org/Labrador/pull/349#issuecomment-3256100441).

mi-hol:
> Thoughts @espotek-org/maintainers ?

There seems to be no feedback. 

@EspoTek 
I'd suggest to merge the PR and in case of unforeseen issues its fast and easy to revert.

EspoTek:
At this stage I think we don't need it, since it's not the kind of software that these types of vulnerability scanners are really looking for.

mi-hol:
What alternative would you see to address the shortage of experienced reviewers?

mi-hol:
@EspoTek with merge of https://github.com/espotek-org/Labrador/pull/368 I see evidence for improving internal processes by using code scanning tools for the project.

All now accepted code changes have been proposed by CodeQL (I have it enabled in my private fork).

Are we loosing anything by enabling code scanning tools for the project?

I don't believe so, instead we can only gain benefits.

EspoTek:
> @EspoTek with merge of #368 I see evidence for improving internal processes by using code scanning tools for the project.
> 
> All now accepted code changes have been proposed by CodeQL (I have it enabled in my private fork).
> 
> Are we loosing anything by enabling code scanning tools for the project?
> 
> I don't believe so, instead we can only gain benefits.

I just fear it could present things as serious security emergencies when they are in fact they are not really possible or likely to ever exploit.

mi-hol:
> I just fear it could present things as serious security emergencies when they are in fact they are not really possible or likely to ever exploit.

In case that happens I'll be the first to disable it. As written my testing showed no signs of this bad results

EspoTek:
Fair, ship it!

PR #368, opened 2025-11-01
add permissions to workflows to enhance security

mi-hol:


EspoTek:
Any reason we've assigned write permissions to PRs?

mi-hol:
@EspoTek from https://medium.com/@morepravin1989/github-actions-security-permissions-a-beginners-guide-4b5f1f86a321

the reason is "pull-requests: write → can update PRs (needed for labeling)"

mi-hol:
and https://shreyapohekar.com/blogs/the-right-way-to-handle-permissions-in-github-actions-a-practical-guide-to-staying-secure/
"**3. Why You Are Still Safe Even With write-all in a pull_request Trigger**
This is the part many people misunderstand.

Even though GitHub allows you to configure:

permissions: write-all
…you are still safe under the pull_request trigger.

Here’s why:

1. GitHub automatically downgrades permissions for workflows triggered by untrusted PRs.
If someone from a fork opens a PR:

The workflow runs in a read-only security sandbox.
Secrets are not available.
GITHUB_TOKEN is forced to read-only, even if you explicitly set it to write-all.
This is why pull_request is considered the safe event."

EspoTek:
I see, so it means that the workflow has permission to write to the PRs, not that people submitting PRs have write access to the repo.

Sounds good, ship it.

Issue #245, opened 2023-05-02
Building with avr-gcc


dbsamithey:
With the aim of compiling the AVR code on Mac OS, I called this script, https://github.com/noctarius/as2make/ , using
```
./as2make --output-name=labrafirm_new --toolchain=/usr/local/avr --toolchain-prefix=avr- --dfp-sdk=./Atmel.XMEGAA_DFP.1.0.39 --project=./USB_BULK_TEST.cproj
```
to generate most of the attached Makefile. 
I had to add the compilation blocks for the assembly files ccp.s and nvm_asm.s manually.  Also, I downloaded that DFP from http://packs.download.atmel.com/ and used packages from https://ftp.gnu.org/gnu/ and http://download.savannah.gnu.org/releases/avr-libc/ to build avr-gcc, avr-binutils, and avr-libc.
Running "make" resulted in "multiple definition" errors complaining about the cali_value_[negative and positive
]_gradient variables that are declared in /src/tiny_calibration.h .  Following advice from stackoverflow, I added the extern keyword to the declarations and added 
```
volatile unsigned char cali_value_positive_gradient;
volatile unsigned char cali_value_negative_gradient;
```
to the top of tiny_calibration.c.
After this change was made, the Makefile ran smoothly and produced the required labrafirm_new.hex output file.  I connected the Labrador to my computer in bootloader mode and ran 
dfu-programmer atxmega32a4u erase --force
dfu-programmer atxmega32a4u flash labrafirm_new.hex
from the command line to upload the .hex, which was successful.
After disconnecting and reconnecting the Labrador, I got an error in the desktop app along the lines of "expected version 2, but seeing version 1."  To address this, I uncommented `#define SINGLE-ENDPOINT-INTERFACE` in globals.h and again compiled and flashed the firmware.  Unfortunately, it still did not function, despite nominally being the same as labrafirm_0006_02.hex.  The best response that I have seen is the brief appearance of an oscilloscope trace, followed by its permanent disappearance, with 'unified debug is of size 27' meanwhile being printed to the console.
[...]

Issue #231, opened 2022-11-08
Calibration has been abandoned due to out-of-range values.


dev-null2019:
Hi!
Compiled the software on Linux(Kali)/ARM64.
Sadly Iam not able to calibrate the hardware - the scope shows a signal at 48V on both channels without anything connected.

On my Windows machine the hardware works without problems.

Regards
Stefan

![2022-11-08 11_13_27-Window](https://user-images.githubusercontent.com/56407056/200537807-3cc33170-b995-42c4-8132-a852fd26e161.png)

Logs:

```
└─$ labrador                                                                                       
QObject::connect: No such slot bufferControl::signalGenIn(bool)
QObject::connect:  (sender name:   'signalGenGroup_CH2')
QObject::connect:  (receiver name: 'bufferDisplay')
QObject::connect: No such slot isoDriver::setWindow(int)
QObject::connect:  (sender name:   'timeBaseSlider')
QObject::connect:  (receiver name: 'controller_iso')
QObject::connect: No such slot timedTickBox::resetTimer(bool)
QObject::connect:  (sender name:   'lockPsuCheckBox')
QObject::connect:  (receiver name: 'lockPsuCheckBox')
Attempting to open "/usr/bin/EspoTek-Labrador/waveforms/_list.wfl"
nameBuffer =  Sin

nameBuffer =  Square

nameBuffer =  Triangle

nameBuffer =  Sawtooth

nameBuffer =  DC
newName =  "Sin"
opening "/usr/bin/EspoTek-Labrador/waveforms/Sin.tlw"
Length =  128
Divisibility =  4
List loaded!!
Attempting to open "/usr/bin/EspoTek-Labrador/waveforms/_list.wfl"
nameBuffer =  Sin

nameBuffer =  Square

nameBuffer =  Triangle

nameBuffer =  Sawtooth

nameBuffer =  DC
newName =  "Sin"
opening "/usr/bin/EspoTek-Labrador/waveforms/Sin.tlw"
Length =  128
Divisibility =  4
List loaded!!
Making USB Driver invisible!!
pipeID 0 = 129
Generic Usb Driver setup complete
unixUsbDriver created!
driver =  unixUsbDriver(0xaaaaf8edd8f0)
axes =  QCustomPlot(0xaaaaf898a9f0, name="scopeAxes")
CHECKING CONNECTION!
Entering unixUsbDriver::usbInit
Libusb context initialised
Device found!!
KERNEL DRIVER INACTIVE
[...]

PR #369, opened 2025-11-09
added Makefile for firmware

brentfpage:
I wrote up a guide for compiling the firmware using gnu command line tools [here](https://github.com/espotek-org/Labrador/wiki/Building-from-source#building-the-firmware) in the wiki.  I think this is a nice ability to enable as it widens the pool of users who are able to customize the firmware.  The build requires a Makefile, which I included in this PR.

### DIFF
--- a/AVR_Code/USB_BULK_TEST/Makefile
+++ b/AVR_Code/USB_BULK_TEST/Makefile
@@ -1,3 +1,4 @@
+# see https://github.com/espotek-org/Labrador/wiki/Building-from-source#building-the-firmware
 EXECUTABLES :=
 LIB_DEP :=
 LIBS :=


mi-hol:
> widens the pool of users who are able to customize the firmware

@brentfpage your efforts to cross platform portability are very much appreciated!
To avoid user confusion, may I ask you add a reference to this PR into https://github.com/espotek-org/Labrador/wiki/Building-from-source#building-the-firmware until it gets merged?

1. I hope the PR can be reviewed and merged quickly
@espotek-org/maintainers any chance someone with the required technical background could own this and run with it quickly?

2. I believe we should fix the reasons for errors outlined in "type make all, and hit enter, which will probably produce errors that can be resolved by making the following changes in USB_BULK_TEST/src" as the second priority.  
Could the changes have negative side effects on Windows Atmel Studio users?

mmehari:
>The only notable difference in the Makefile included in this repo relative to that produced by as2make is -funsigned-char -> -fsigned-char

Can you please add the reason for the change?

I also tested on my system (Ubuntu 24.04) and it compiled without a problem. I wonder on which system you have tried and encountered the errors.

brentfpage:
> To avoid user confusion, may I ask you add a reference to this PR into https://github.com/espotek-org/Labrador/wiki/Building-from-source#building-the-firmware until it gets merged?

@mi-hol Good idea, done now


> I believe we should fix the reasons for errors outlined...

A few more of the edits can be contextualized a bit:

- stuff related to `cali_value_(posi/nega)tive_gradient`
   - Without the described edits, I get "multiple definition" errors for these variables in every file that `#include`s tiny_calibration.h, which judging from [this stackoverflow thread](https://stackoverflow.com/questions/11072244/c-multiple-definitions-of-a-variable) is a commonly-encountered issue.  I made the fix suggested in that thread.
- comment-out `udd_ctrl_request_t udd_g_ctrlreq;` on line 168 of main.c
   - an alternative is to prepend `extern` to the repeated definition of this variable on line 252 of src/ASF/xmega/drivers/usb/usb_device.c

I'm slightly more worried about the `udd_set_setup_payload` casting edits.  Casting `&firmver` to `(uint8_t *)` could conceivably be problematic because `firmver` is actually an `unsigned short`, i.e., 16 bits.  But, there are other places where similar casts are made in `udd_set_setup_payload` calls, e.g. on lines 436-437 of src/ASF/common/services/usb/udc/udc.c,
```
	udd_set_setup_payload( (uint8_t *) & udc_device_status,
			sizeof(udc_device_status));
```
where `udc_device_status` has type `uint16_t` (a.k.a. `le16_t`).  All things considered, the casting edits I'm almost certain are harmless.

> Can you please add the reason for the change? (-funsigned-char -> fsigned-char)

@mmehari In my experience (described in #245), the firmware compiles but does not run properly when `-funsigned-char` is used.  I added a link to #245 to the wiki.  Using `-fsigned-char` was originally suggested in #231, albeit in the context of the software interface.  Maybe I just got lucky that this setting is relevant to the firmware as well, but it makes the difference between near-0% functionality and seemingly 100%.

mmehari:
I remember using the Atmel Studio (MPLab X) export utility to generate a Makefile project. That way I was able to create a new firmware image (under Linux) for https://github.com/espotek-org/Labrador/pull/263.

Unfortunately, I didn't save the Makefile project and it could have been nice to compare the two.

Nevertheless, I like a dedicated Makefile being present for firmware development.

brentfpage:
@mmehari And I guess you didn't encounter the need to make small edits like this to the source code?  Maybe my impression that gcc versioning is responsible for them is wrong.  But it seems like Atmel/Microchip does in fact use the same GNU tools for firmware compilation (https://www.microchip.com/en-us/tools-resources/develop/microchip-studio/gcc-compilers), so I'm at a loss as to the origin of the differences in compiler complaints.

mmehari:
Indeed I didn't encounter to make the changes by then but also this time. I would expect a similar behavior had I tried compiling my Makefile project on your platform. So I will rule out such a concern. But I wonder on which platforms does the compilation fails

brentfpage:
Oh right makes sense, it's a platform-by-platform thing.  I'm personally on a mac with an intel processor.  So we know that running the makefile without source code edits works on ubuntu (maybe all linux systems?) and with source code edits works on an intel mac.

I would be extremely surprised if the requirement to use ~~`-funsigned-char`~~ `-fsigned-char`(edited, got this wrong before) is platform-dependent though as I think the Makefile gets run essentially as written, without optimization and preprocessing steps as for the source code before compilation.

mmehari:
I confirm that the firmware compiled with the `-funsigned-char` Makefile flag doesn't work on my system (Ubuntu 24.04) and I guess that is platform independent.

mi-hol:
> I confirm that the firmware compiled with the `-fsigned-char` Makefile flag doesn't work on my system (Ubuntu 24.04) and I guess that is platform independent.

@mmehari would it be correct to say: removing -fsigned-char would allow us to merge the PR?

mmehari:
No need to change the PR. What I was checking was whether the original Makefile (`-funsigned-char`) from as2make works on my system (Ubuntu 24.04). Since it doesn't work, the update from `-funsigned-char` to `-fsigned-char` was necessary as was originally done by @brentfpage.

So no need to change the PR and we can merge it as is.

mi-hol:
@mmehari out of curiosity, are we using the makefile for our build actions?
in case not, would we benefit from using it?

mmehari:
We don't have build action for the firmware project and it should be possible to use the Makefile for this purpose. However, firmware development is rarely done and I don't see the importance of having one

mi-hol:
> However, firmware development is rarely done and I don't see the importance of having one

Thanks Michael, please allow me to have a different view on this due to reasons layed out in #354  & https://github.com/espotek-org/Labrador/pull/349#issuecomment-3256100441

mmehari:
I see your point of view and indeed that makes sense to avoid surprise changes in the firmware. I was just looking into the firmware commits which weren't that many changes.

PR #370, opened 2025-11-11
arm64-v8a build

brentfpage:
The unsigned arm64-v8a apk should be in /Desktop_Interface/android-build//build/outputs/apk/release/android-build-release-unsigned.apk.

I also added the following 64 bit libraries:
- liblog from https://github.com/EdgeApp/android-ndk-linux/blob/master/platforms/android-24/arch-arm64/usr/lib/liblog.so
- libusb from https://github.com/louisgeek/LG_libusb_Android/tree/main/arm64-v8a
- libdfuprog from the 'adjustments-for-arm64_v8a' branch  of my fork of the libdfuprog repo (www.github.com/brentfpage/libdfuprog/)

I haven't looked into the specific libusb features that were implemented in the custom armeabi-v7a version of the library and whether or not they're included in this new 64 bit version.  I'm also not 100% confident in the libdfuprog build because I had to grab from various sources some deprecated android ndk libraries that may not play nice together.  But, everything compiled smoothly, so this PR I hope is forward progress on an arm64-v8a build.

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -60,13 +60,13 @@ jobs:
             qmake -config release
             make -j$(nproc)
             make INSTALL_ROOT=android-build install
-            JAVA_HOME=${JAVA_HOME_8_X64} androiddeployqt --input android-libLabrador.so-deployment-settings.json --output android-build --aab --apk --android-platform android-31 --verbose --gradle --release
+            JAVA_HOME=${JAVA_HOME_8_X64} androiddeployqt --input android-Labrador-deployment-settings.json --output android-build --aab --apk --android-platform android-31 --verbose --gradle --release
             ${ANDROID_SDK_ROOT}/build-tools/31.0.0/apksigner sign --in android-build/build/outputs/apk/release/android-build-release-unsigned.apk --out ${APK_BASENAME}-release-signed.apk --key cert.p8 --cert cert.cer --verbose
           else
             qmake -config debug
             make -j$(nproc)
             make INSTALL_ROOT=android-build install
-            JAVA_HOME=${JAVA_HOME_8_X64} androiddeployqt --input android-libLabrador.so-deployment-settings.json --output android-build --aab -apk --android-platform android-31 --verbose --gradle
+            JAVA_HOME=${JAVA_HOME_8_X64} androiddeployqt --input android-Labrador-deployment-settings.json --output android-build --aab -apk --android-platform android-31 --verbose --gradle
             cp android-build/build/outputs/apk/debug/android-build-debug.apk ${APK_BASENAME}-debug.apk
           fi
           cp android-build/build/outputs/bundle/release/android-build-release.aab ${APK_BASENAME}-release-unsigned.aab


brentfpage:
In commits 00d6720 - b234865, I update the target SDK to 31.  A GH runner action on the last of these commits, b234865, produces a functional arm64-v8a app for my Pixel 6A running Android 16.  Apart from bumping up version numbers in Labrador.pro, android.yml, and AndroidManifest.xml, the only changes required for the SDK update were:

- setting `android:exported="true"` in AndroidManifest.xml:
    - Reason: from https://developer.android.com/guide/topics/manifest/activity-element#exported:
"If an activity in your app includes intent filters, set this element to "true" to let other apps start it. For example, if the activity is the main activity of the app and includes the category android.intent.category.LAUNCHER."

- adding a FLAG_MUTABLE argument to a PendingIntent constructor in androidInterface.java
    - Reason: without this flag, the app instacrashes if SDK=31 is targeted.  From [here](https://developer.android.com/reference/android/app/PendingIntent#FLAG_MUTABLE), "up until (Android 11/SDK 30) PendingIntents are assumed to be mutable by default" . To be conservative, I maintained the PendingIntent as mutable by adding FLAG_MUTABLE.  Also, using code from [here](https://stackoverflow.com/questions/70894168/targeting-s-version-31-and-above-requires-that-one-of-flag-immutable-or-flag), I added some checking of the SDK version so that this flag (which was introduced for SDK 31) is not passed as an argument for SDK<31.

EspoTek:
Looking good man.  Does the "android" runner produce v7a as well as v8a, or do we need a separate workflow for that?

brentfpage:
Thanks a lot!  The one "android" runner builds for both architectures, and it seems that these are packaged into a single apk by default.  The apk gets uploaded by the final `Upload apk artifacts` block as previously.

I also added `CONFIG += android_app_bundle` to Labrador.pro and an `--aab` flag to androiddeployqt so that an .aab file is produced.  This was done in case the Google Play store requires an .aab nowadays, but it could be unnecessary.  Also, I haven't worked out signing support for the aab yet.

EspoTek:
That's awesome man.  We ready to merge?  I just looked over the code and it all seemed good!

brentfpage:
Thanks!  Yeah I was relieved to see that building for multiple archs is fairly streamlined for qt >= 5.14 .  We are ready to merge!

brentfpage:
Just saw that the build failed, which is very surprising considering that the same workflow on the same commit worked in the fork, albeit for the unsigned debug version. edit: will follow up in #343

EspoTek:
Yeah, I have no idea either.  Currently running it through the AI to see if there's some obvious fix, if not we might need to dig deeper!

brentfpage:
Think I might have some found some errant --apk flags from me in android.yml!  Trying a runner action without them. Funnily enough, an extra typo from me in the debug version block "--apk -> -apk" might have made it work before.

brentfpage:
Hopefully not too rashly merged a small tweak to android.yml in the genuine master branch.  Thanks a lot for adding me as a maintainer!  Given that the debug version works, I think there's a good chance that the release version will work as well now with this new commit, but don't want to use up GH runner capital unless you agree.

EspoTek:
> Hopefully not too rashly merged a small tweak to android.yml in the genuine master branch. Thanks a lot for adding me as a maintainer! Given that the debug version works, I think there's a good chance that the release version will work as well now with this new commit, but don't want to use up GH runner capital unless you agree.

Easy.  Feel free to commit small stuff like this to master without review, but I would still recommend creating a new branch and testing there before doing so:

```
git checkout -b my_new_branch_name
git push origin my_new_branch_name
```

You can run github actions against the custom branches before then taking your commits over to master (I use `git cherry-pick` for this).

brentfpage:
OK got it!  Also, `git cherry-pick` looks quite versatile and I'll have to try it out.

I'm relieved to see that the android build now works.  Apologies for the 11th hour change – it's hard to know when to stop testing and this time I should've gone just a bit further.

EspoTek:
All good man, seriously appreciate everything.

If you're new to git, the recommended "advanced beginner" commands to learn ASAP would be:
 - `git stash` (always a better option than `git reset --hard HEAD`), including `git stash apply` (always better than `git stash pop`), `git stash show`, `git stash show -p`
 - `git cherry-pick`
 - `git bisect` (very handy for debugging)
 - `git reflog` (can save you from pretty much any situation, ever)
 - `git blame`
 also:
 - `git pull --rebase`

brentfpage:
Not at all!  Productive learning-by-doing is my favorite kind of activity.

Those git tips are much appreciated.  I need to up my `stash` game – from those examples it seems like it functions almost as a whole new mini version control system within git!

PR #375, opened 2025-12-10
Avoid blank dialog boxes that halt calibration on android

brentfpage:
I kept on encountering blank dialog boxes like the one below when going through the cal' sequence on android.
Screenshot_20251210-143000
In this PR, the calibrationMessages dialog box object gets reinitialized before every message.  While not very elegant, this change seems to prevent the blank dialog boxes.

I also added
```
        ui->controller_iso->driver->psu_offset = psu_voltage_calibration_offset;

        ui->controller_iso->driver->calibrateOnConnect = dt_userWantsToCalibrate;
        connect(ui->controller_iso->driver, SIGNAL(calibrateMe()), this, SLOT(on_actionCalibrate_triggered()));
```
to the end of the PLATFORM_ANDROID block in the MainWindow constructor.  It seems like this is required for consistency with what is done for the other platforms, i.e., what is done in the function reinitUsb() and reinitUsbStage2().  I'll test the effect of this change a bit more later and remove it if it appears unnecessary. (edit: the lines involving calibration do appear necessary and improve performance.  Also, the line involving the psu_offset I'm fairly sure is required for loading in calibration data in case the user has already calibrated.)


brentfpage:
Yikes don't know why this is 28 commits.  I figured that merging from genuine/master would sync the fork up, but maybe different commit hashes are used in the fork?  Anyway, if/when these changes are incorporated, I'll do it in a bit of a cleaner way than through this PR.

EspoTek:
> Yikes don't know why this is 28 commits.

In this situation, the simplest solution (that I know of) is to do something like:
```
git diff master HEAD > patch.diff
git reset --hard master 
git apply patch.diff
git commit -a -m "Commit message goes here"
```

If you screw something up, you can use `git reflog` to find the commit hash from before you reset.

EspoTek:
That said I'm basically happy to get this merged.

But first, I think we need to
 - Remove the .apk from git (this shouldn't be tracked)
 - Ensure that this didn't break the calibration on desktop

brentfpage:
> In this situation, the simplest solution (that I know of) is to do something like: 
```
git diff master HEAD > patch.diff
git reset --hard master 
git apply patch.diff
git commit -a -m "Commit message goes here"
```

Ahh that's satisfying

brentfpage:
> But first, I think we need to
> - Remove the .apk from git (this shouldn't be tracked)
  - will do.  Would `git rm ...` be best?
> - Ensure that this didn't break the calibration on desktop
  - I tested the calibration on Mac with this PR's changes incorporated and it worked well.  If testing on windows/linux is also necessary, I unfortunately won't be able to do it because I don't have access to such machines currently.

One last thing that I'll try to work out is making sure that if an android user connects the board, opens the app, then agrees to calibrate in response to the pop-up, that calibration does then get initiated as desired.  Right now, it doesn't, I think because the board has already been connected when the user agrees to calibrate.  On mac, the same sequence DOES initiate calibration, so there's probably an easy fix.

brentfpage:
> One last thing that I'll try to work out is making sure that if an android user connects the board, opens the app, then agrees to calibrate in response to the pop-up, that calibration does then get initiated as desired.

To do this, I added 

```
if((ui->controller_iso->driver->connected)&&dt_userWantsToCalibrate)
  QTimer::singleShot(0, this, SLOT(on_actionCalibrate_triggered()));
```
to the `PLATFORM_ANDROID` block in the MainWindow constructor.  From [this stackoverflow post](https://stackoverflow.com/questions/2356778/closing-a-qmainwindow-on-startup), I learned that timers only start after the main event loop has started.  This happens after the MainWindow has been constructed (as shown in `main.cpp`), so the singleShot above functions essentially as a callback upon completion of the UI setup.  I tested the code and it worked well.

brentfpage:
Closing and will reopen on an equivalent new branch apart from cleaner file history

Issue #260, opened 2024-03-14
Extending Labrador vendor requests


mmehari:
Hello again,

This time I am planning to add a new feature which requires a new vendor request but while experimenting to add one, I ended up having the following error message

> unixUsbDriver::usbSendControl FAILED with error LIBUSB_ERROR_PIPE

What I did to recreate the error message was updated SIGNAL GEN CH1 command from **0xa1** to **0xb1** in the following two files
1. https://github.com/EspoTek/Labrador/blob/master/AVR_Code/USB_BULK_TEST/src/ASF/common/services/usb/udc/udc.c#L1108
2. https://github.com/EspoTek/Labrador/blob/master/Desktop_Interface/genericusbdriver.cpp#L235

Would you please let me know what I am missing?

EspoTek:
Hi again, Michael!

At least in theory, that should have done the trick!

In practice, though, there's a feature that automatically flashes the "correct" firmware to the board that could be interfering with your code.  To rule this out, it'd be best practice to increment the `EXPECTED_FIRMWARE_VERSION` in the desktop app, and the `FIRMWARE_VERSION_ID` in globals.h on the microcontroller side.

If you're running on any operating system that isn't Windows x64, you'll also want to uncomment the line defining `SINGLE_ENDPOINT_INTERFACE` (also in globals.h).  This changes the headers so that it uses 1x 1023-byte isochronous endpoint, rather than 6x 128-byte endpoints to send the scope/logic analyzer data.  On non-Windows-x64, the ordering of the transfers isn't deterministic, so using 6x endpoints leads to a corrupt signal on the oscilloscope/logic analyzer.

The other thing I'll mention in general is that Wireshark is an excellent tool for capturing raw(ish) USB packet data.  You can see exactly what data is sent to the device, and exactly what the response is.  Just make sure you check the "install USBPcap" box when installing it if you're on Windows (don't install USBPcap separately - it doesn't work properly).

mmehari:
Indeed you were right.

My firmware was overwritten by the default firmware and increasing the FIRMWARE_VERSION_ID solves the issue.

And yes I am already using Wireshark to debug the low level USB traffic and indeed it is very helpful.

Now I will focus on the new feature I am trying to bring and hopefully, I will see you in a new pull request

EspoTek:
Great news!  Can't wait to see the PR!  I'll mark this issue as closed now.

PR #381, opened 2025-12-18
'scope firmware DMA settings edits

brentfpage:
These edits to the DMA settings in the firmware might improve the performance of #367 at frequencies above 25 khz, but it's not obvious that they do.
Changes like those to line 17,
`DMA.CH0.ADDRCTRL = DMA_CH_SRCRELOAD_BURST_gc | DMA_CH_SRCDIR_INC_gc | DMA_CH_DESTDIR_INC_gc | DMA_CH_DESTRELOAD_BLOCK_gc;   //Source reloads after each burst, with byte incrementing.  Dest does not reload, but does increment address.`
->
`DMA.CH0.ADDRCTRL = DMA_CH_SRCDIR_FIXED_gc | DMA_CH_DESTDIR_INC_gc | DMA_CH_DESTRELOAD_BLOCK_gc;   //Source reloads after each burst, with byte incrementing.  Dest does not reload, but does increment address.`
may just be a matter of taste, but it seems possible that the second method takes fewer steps.  From what I can tell, these transfers are both 1 byte per burst and 1 byte per transfer, so 1 transfer per burst.  If `DMA_CH_SRCRELOAD_BURST_gc | DMA_CH_SRCDIR_INC_gc` is set, the source address gets incremented after a transfer, but because that is equal to a burst, it then gets immediately set to the original address.  This might be achieved more directly using `DMA_CH_SRCDIR_FIXED_gc`.

I'm more confident in the change of the DMA CH1 trigger source to `DMA_CH_TRIGSRC_ADCA_CH2_gc` in `tiny_dma_set_mode_2`.  In the current firmware, it's instead set to `...ADCA_CH0_gc`, but the source address for DMA CH1 is ADCA_CH2.  It appears this discrepancy almost never or never matters because both ADC_CH0 and ADC_CH2 are sampling at the same frequency in this mode, but maybe it rears its head in some scenarios.

### DIFF
--- a/AVR_Code/USB_BULK_TEST/src/tiny_dma.c
+++ b/AVR_Code/USB_BULK_TEST/src/tiny_dma.c
@@ -284,7 +284,7 @@ void tiny_dma_set_mode_2(void){
 	DMA.CH3.DESTADDR2 = 0x00;
 	
 	//Must enable last for REPCNT won't work!
-	DMA.CH3.CTRLA |= DMA_CH_ENABLE_bm;  //Enable!
+// 	DMA.CH3.CTRLA |= DMA_CH_ENABLE_bm;  //Enable!
 	
 	DMA.CH0.CTRLA = 0x00;
 	DMA.CH0.CTRLA = DMA_CH_RESET_bm;
@@ -307,7 +307,7 @@ void tiny_dma_set_mode_2(void){
 	DMA.CH1.CTRLA = DMA_STANDARD_CTRLA;
 	DMA.CH1.CTRLB = DMA_STANDARD_INTERRUPT;
 	DMA.CH1.ADDRCTRL = DMA_CH_SRCRELOAD_BURST_gc | DMA_CH_SRCDIR_INC_gc | DMA_CH_DESTDIR_INC_gc | DMA_CH_DESTRELOAD_BLOCK_gc;   //Source reloads after each burst, with byte incrementing.  Dest does not reload, but does increment address.
-	DMA.CH1.TRIGSRC = DMA_CH_TRIGSRC_ADCA_CH0_gc;	//Triggered from ADCA channel 0
+	DMA.CH1.TRIGSRC = DMA_CH_TRIGSRC_ADCA_CH2_gc;	//Triggered from ADCA channel 2
 	DMA.CH1.TRFCNT = DMA_STANDARD_TRANSFER_LENGTH;
 				
 	DMA.CH1.SRCADDR0 = (( (uint16_t) &ADCA.CH2.RESL) >> 0) & 0xFF; //Source address is ADC


mi-hol:
The questions in https://github.com/espotek-org/Labrador/pull/385#issuecomment-3679454853 are linked with merging this PR

mi-hol:
The comment from https://github.com/espotek-org/Labrador/issues/260#issuecomment-2000606393
would seem to apply for this PR too.

> My firmware was overwritten by the default firmware and increasing the FIRMWARE_VERSION_ID solves the issue.

mi-hol:
@brentfpage did you see above comment https://github.com/espotek-org/Labrador/pull/381#issuecomment-3679518122?
I miss an update for FIRMWARE_VERSION_ID  in the changed files :)

brentfpage:
@mi-hol Good point, done now

mi-hol:
The log shows no files named labrafirm_0008*  as I would have expected.

1. Is my expectation maybe wrong
**OR**
2. Are we missing a "build firmware" action or other/additional changes?


image

brentfpage:
I don't think there's an action for building the firmware.

mi-hol:
> action for building the firmware.

@brentfpage creating this missing workflow, would that be a challenge you'd like to own?

brentfpage:
> > action for building the firmware.
> 
> @brentfpage creating this missing workflow, would that be a challenge you'd like to own?

Started a discussion in #396

mi-hol:
> Started a discussion in #396

@brentfpage from [your comment there](https://github.com/espotek-org/Labrador/discussions/396#discussioncomment-15534681) I assume  a new PR for an action that builds and uploads 2 firmware  file to [Labrador firwmare directory](https://github.com/espotek-org/Labrador/tree/master/Desktop_Interface/resources/firmware/).  
Is that correct understanding?
This new PR would then unblock merging this PR

brentfpage:
> > Started a discussion in #396
> 
> @brentfpage from [your comment there](https://github.com/espotek-org/Labrador/discussions/396#discussioncomment-15534681) I assume a new PR for an action that builds and uploads 2 firmware file to [Labrador firwmare directory](https://github.com/espotek-org/Labrador/tree/master/Desktop_Interface/resources/firmware/). Is that correct understanding? This new PR would then unblock merging this PR

That sounds good.  I'm not sure what the best way to do it is though.  Is there a way to modify the Labrador repo using a build action– specifically, to upload firmware files to [this directory](https://github.com/espotek-org/Labrador/tree/master/Desktop_Interface/resources/firmware) in this case?  From what I've gathered, the "actions/checkout@v4" line at the top of all the workflow files actually clones the repo, so there's no easy access to the genuine repo.

Also, it would be nice to have some sort of automatic versioning so that FIRMWARE_VERSION_ID in AVR_Code/..../globals.h and EXPECTED_FIRMWARE_VERSION in Desktop_Interface/genericusbdriver.h get incremented when the avr code is updated.  I hope to add that functionality.

mi-hol:
> I'm not sure what the best way to do it is though. Is there a way to modify the Labrador repo using a build action– specifically, to upload firmware files to [this directory](https://github.com/espotek-org/Labrador/tree/master/Desktop_Interface/resources/firmware) in this case?

**updated:**
Uploading firmware files to [this directory](https://github.com/espotek-org/Labrador/tree/master/Desktop_Interface/resources/firmware) is the correct solution.
An example I know is on https://github.com/FOME-Tech/fome-fw/blob/master/.github/workflows/build-firmware.yaml
Does that help?

mi-hol:
> it would be nice to have some sort of automatic (firmware) versioning

**updated**:
Considering the currently existing pain points of firmware versioning, automating versioning
 would have lower priority from my view and I'd recommend to address it only after the first part (above) was achieved.
Main reason is that a better solution than currently used "file naming approach" is required from my view.

brentfpage:
> Uploading firmware files to [this directory](https://github.com/espotek-org/Labrador/tree/master/Desktop_Interface/resources/firmware) is the correct solution.
> An example I know is on https://github.com/FOME-Tech/fome-fw/blob/master/.github/workflows/build-firmware.yaml

Maybe I'm missing something, but doesn't
```
    - name: Upload ini
      uses: actions/upload-artifact@v4
      with:
        name: ${{matrix.ini-file}}
        path: ./firmware/tunerstudio/generated/${{matrix.ini-file}}
```
in that workflow upload _from_ `./firmware/tunerstudio/generated/`, not to it?  I see that `/firmware/tunerstudio/generated/` in the genuine repo does actually contain firmware files, but I'm guessing they're added to it separately for the following reasons:

- In Labrador mac.yml,
```
      - name: Upload artifacts
        if: ( env.MACOS_CERTIFICATE != '' && env.MACOS_CERTIFICATE_PWD != '' )  
        uses: actions/upload-artifact@v4
        with:
          name: asset-dmg
          path: Desktop_Interface/Labrador*.dmg
```
uploads `Desktop_Interface/Labrador*.dmg` from the workflow's clone of the repo.   The file `Desktop_Interface/Labrador*.dmg` never appears in the genuine repo.

- [this](https://www.codestudy.net/blog/is-there-a-way-to-push-changes-with-a-github-action/) suggests that pushing changes from the workflow clone to the genuine repo requires doing something like configuring personal access tokens (PAT) and adding references to them in the workflow file

To clarify, I think the following happens:
- when `build-firmware.yaml` is run:
  - firmware files are installed in `./firmware/tunerstudio/generated/` in the workflow's clone of the repo
  - `upload_artifacts` uploads these to a directory accessible to the developers, similarly to how `upload_artifacts` in mac.yml results in Labrador.dmg being downloadable via a link provided by github
  - the workflow finishes, the clone of the repo is deleted, and ./firmware/tunerstudio/generated/ in the genuine repo has not been touched
- some other mechanism later uploads the artifacts into `./firmware/tunerstudio/generated/` in the genuine repo

mi-hol:
> uploads `Desktop_Interface/Labrador*.dmg` from the workflow's clone of the repo. The file `Desktop_Interface/Labrador*.dmg` never appears in the genuine repo.
> 
> * [this](https://www.codestudy.net/blog/is-there-a-way-to-push-changes-with-a-github-action/) suggests that pushing changes from the workflow clone to the genuine repo requires doing something like configuring personal access tokens (PAT) and adding references to them in the workflow file


**Updated:**
Maybe we should break with manual legacy process for firmware and stop uploading compiled firmware files to the repo. 
That change would solve the issue of obsolete, orphaned firmware files in repo too!


Maybe the better place for handling the new process would be in https://github.com/espotek-org/Labrador/blob/master/.github/workflows/continuous.yml  ?
@turboencabulator maybe, as the author of continuous.yml, you'd have a recommendation ?

turboencabulator:
I don't currently have any recommendations, but I have lots of thoughts and opinions!

Putting generated files in a repo:  Almost always a bad idea.  In my experience some generated files aren't binary reproducible, they've got embedded timestamps or paths, or the output depends on the exact version of your compiler or OS.  Alice commits a generated file, Bob clones the repo and builds it, and now Bob sees the generated file in a modified state.  Maybe Bob ignores it, or maybe he commits it and now Alice has that same problem.  These are almost always large binary files that bloat the repo.  My .git directory for this project is currently about 140MB because even after you delete a file, git still has to remember it.

Automated commits (either to commit firmware files or to bump all the version numbers):  Probably more trouble than it's worth, and there are more ways for it to go wrong than to go right.  I know there is a lot of trial and error to get the automation working correctly, while firmware isn't going to see many updates to make the cost vs. benefit worth it.  I'd suggest documenting the instructions somewhere and using that as a checklist for how to release the firmware in the future.

Just like we don't need to run the release process on every commit, I doubt we need to bump the firmware version numbers on every commit that touches firmware.  My preference is to increment it when there is a new feature or improvement, and after we think it's stable (basically the same reasons to run the `continuous` workflow).  The problem is that the desktop software forces you to use a particular firmware version, making proposed firmware changes hard to test.

We're getting pretty off-topic for a PR originally about DMA settings.  If those DMA changes work, then let's do it and move this discussion elsewhere for what we want the process to look like.

mi-hol:
> I doubt we need to bump the firmware version numbers on every commit that touches firmware. 

@turboencabulator  According to Chris's comment https://github.com/espotek-org/Labrador/issues/260#issuecomment-2000606393, firmware gets only applied to hardware when the firmware version number is incremented. So this is a fact!

> If those DMA changes work, then let's do it and move this discussion elsewhere for what we want the process to look like.


I'd be happy to do as you suggest, unfortunately to test for Windows, the installer config ([Desktop_Interface/build_win/Labrador.aip](https://github.com/espotek-org/Labrador/blob/c9dd7f2e2bf2a0b17dcbb33e287f159aed5df614/Desktop_Interface/build_win/Labrador.aip#L37)) will need to be modified. Not sure about about the LINUX installers because searching for "firmware" and "labrafirm" had no hits in the related "build_*" dirs.
Could you comment please?


image

mi-hol:
Note: related PR https://github.com/espotek-org/Labrador/pull/402 should be merged first

turboencabulator:
> @turboencabulator According to Chris's comment [#260 (comment)](https://github.com/espotek-org/Labrador/issues/260#issuecomment-2000606393), firmware gets only applied to hardware when the firmware version number is incremented. So this is a fact!

Firmware gets automatically loaded whenever the version currently on the hardware doesn't match whatever the desktop software expects.
https://github.com/espotek-org/Labrador/blob/c9dd7f2e2bf2a0b17dcbb33e287f159aed5df614/Desktop_Interface/genericusbdriver.cpp#L484-L491

Or you can fire up dfu-programmer and load what you want, but you have to watch out for this automatic loading to overwrite it.

The point I was trying to make earlier is that this change doesn't break compatibility with the desktop software (at least I don't think it does).  If users won't see any difference, do we really need to make a new firmware version right now and force the desktop software to load that version?

turboencabulator:
> I'd be happy to do as you suggest, unfortunately to test for Windows, the installer config ([Desktop_Interface/build_win/Labrador.aip](https://github.com/espotek-org/Labrador/blob/c9dd7f2e2bf2a0b17dcbb33e287f159aed5df614/Desktop_Interface/build_win/Labrador.aip#L37)) will need to be modified. Not sure about about the LINUX installers because searching for "firmware" and "labrafirm" had no hits in the related "build_*" dirs. Could you comment please?

Check Desktop_Interface/Labrador.pro.  Usually all the firmware files in the directory get copied into the package (no need to list them individually), except Android which only copies a single firmware file to make the package size smaller.

mi-hol:
> that this change doesn't break compatibility

Sorry, I fail to understand this reasoning.
A change was made in firmware and needs to be tested, due to automatic firmware update, making a new firmware version, seems the only approach to reliably deploy it.
It's all about simplicity and avoiding unexpected side effects.

brentfpage:
The changes in this PR now ensure that two-channel scope mode features no timing glitches.  Here's my test setup:
- put the Labrador in two-channel scope mode
- set sig. gen. channel 1 to produce a 58.823 khz waveform
    - the waveform that gets generated has 16 samples and has a sample update period of 51 clock cycles.  (aside: the clock freq. is 48 Mhz)
- input sig. gen Ch. 1 into scope Ch. 1
- use the DAQ feature to collect a five-second data stream from CH. 1 at the maximum DAQ sample rate: 187.5 khz
- figure out the periodicity of the waveform in the data:
    - clock cycles per DAQ sample: 48e6/187.5e3‎ = 256
    - clock cycles per waveform period: 51*16‎ = 816
    - DAQ samples per waveform period = 816/256=3.1875
        - not an integer, so won't work
    - DAQ samples per 16 waveform periods = 51
    - so, the data stream should repeat every 16 waveform periods, or 51 DAQ samples
- lop off the first 10% of the data stream (don't know why this is necessary)
- split the data stream into chunks of 51 samples, plot all the chunks on top of one another

With this PR's changes:
f59_w_patch


Without this PR's changes (the waveform shifts by a DAQ sample every ~0.5 seconds or so):
f59_wo_patch

Unfortunately, this PR (a draft) disables sig. gen. CH2 for the two-channel scope mode.  It seems like there must be a better solution, because none of the xmega's specs are being exceeded.  I think the issue can't be related to the ADC or DAC specifically, because either disabling scope CH2 or disabling sig. gen. CH2 can resolve it.  On the other hand, I've read that DMA can handle something like a sample on each channel every 2 clock cycles, which is way more than is being used, so DMA also doesn't seem to be a likely culprit.

Issue #378, opened 2025-12-16
update ReadMe.md section "Building from Source"


mi-hol:
**Is your feature request related to a problem? Please describe.**
sync doc with enhancement made in https://github.com/espotek-org/Labrador/commit/4e71cc9e6e4a5edaa32ad160f77edd33a69e1eef

**Describe the solution you'd like**
reference new Makefile AVR_Code/USB_BULK_TEST/Makefile

@brentfpage may I kindly ask you to own this task?

brentfpage:
@mi-hol Will do

brentfpage:
I included a draft **Building from Source** section below with bold script highlighting new/revised material.  I ended up mostly making edits to the summary of the desktop interface build and also added instructions for the mac build to the wiki.  

For the Readme summary of the Makefile-based firmware build, I think the new link to the wiki instructions is adequate.  This keeps the Readme as streamlined as possible and tailored to a typical user while giving the Makefile-based firmware compilation guide some exposure.

## Building from Source
If you're looking to build from source but don't know where to start, Qt Creator is the easiest way to get your toes wet!  
https://www.qt.io/download-open-source/  
When installing, make sure you tick the box to install Qt 5.15 or later.

Once it's installed, open `Desktop_Interface/Labrador.pro`, then Clean All -> Run `qmake` -> Build All.

**Alternatively, you can build on the command line.  The process will consist of a subset of the commands present in the build scripts in the `.github/workflows/` directory of this repo.  These are run by Github to produce the packages on the [releases page](https://github.com/espotek-org/Labrador/releases), but include app deployment commands that can be skipped for individual users.**

**Streamlined build scripts for individual Linux (including Raspberry Pi) and macOS users are described in [Building the desktop interface](https://github.com/espotek-org/Labrador/wiki/Building-from-source#building-the-desktop-interface) section of the wiki.  As shown there, following installation of prerequisite packages, the Linux build simplifies to the following sequence of commands run within the `/Desktop_Interface` directory,**
```
QT_SELECT=qt5 qmake
make
sudo make install
sudo ldconfig
```
after which the command `labrador` launches the app.

[...]

Issue #384, opened 2025-12-20
Add visibility of firmware version check & update to app (as in LabraScope)


mi-hol:
**Is your feature request related to a problem? Please describe.**
devices with old firmware are neither identified nor updated

**Describe the solution you'd like**
Handle it automatically at start-up like in LabraScope 

<img width="482" height="222" alt="Image" src="https://github.com/user-attachments/assets/8715fe79-4641-4af4-a3bb-b9ce4863ab21" />

EspoTek:
The regular software already automatically detects and updates (or downgrades, or changes variant of) the firmware automatically.

It just doesn't give a prompt like this.

mi-hol:
> The regular software already automatically detects and updates (or downgrades, or changes variant of) the firmware automatically

That seems not to work correctly then. I used the specific device (Backer's Edition Batch2) since about a year. I verified with a second device (from same batch owned by a friend) with the same result.

mi-hol:
@EspoTek wiki currently only mentions how dfu-programmer can be used to flash firmware.
https://github.com/espotek-org/Labrador/wiki/Troubleshooting-Guide#windows

triggered by "changes variant of the firmware" made me re-read above wiki and I noticed

*_01 is mentioned for Windows and *_02 for UNIX-based OSes.

Are there 2 different "types" of Labrador firmware required?

"The regular software already automatically detects and updates (or downgrades, or changes variant of) the firmware automatically" should be added because it is an important fact.

mi-hol:
note to myself, https://github.com/espotek-org/Labrador/issues/260#issue-2187113092 mentions this too but visibility to user is missing and has already caused confusions to contributors.

from a UNIX Labrador.log file:
```
CHECKING CONNECTION!
Entering unixUsbDriver::usbInit
Libusb context initialised
Device found!!
KERNEL DRIVER INACTIVE
Interface claimed!
Connected
BOARD IS RUNNING FIRMWARE VERSION 0x0007
EXPECTING FIRMWARE VERSION 0x0007
FIRMWARE VARIANT = 0x02
EXPECTED VARIANT = 0x02
Connecting now!

```

PR #385, opened 2025-12-21
"Building AVR Firmware" changes to makefile and documentation

brentfpage:
Doing #378 .  I kept the focus of the "Building from source" section of the Readme on the desktop interface, but added pointers to the firmware build instructions and Makefile to give them some exposure.

### DIFF
--- a/AVR_Code/USB_BULK_TEST/Makefile
+++ b/AVR_Code/USB_BULK_TEST/Makefile
@@ -8,14 +8,15 @@
 # After running 'make 0x01', to switch to building variant 0x02 instead, first run 'make clean'.
 # The same goes for switching from building the 0x02 variant to building the 0x01 variant.
 # Some details:
-# 	In the build process for variant 0x01, the macro SINGLE_ENDPOINT_INTERFACE is
-# 	undefined, while in the build process for variant 0x02,
-# 	SINGLE_ENDPOINT_INTERFACE is defined.  In several regions of the source code,
-# 	the undefined/defined status of this macro determines which of two possible
-# 	code blocks are compiled into the firmware.  The collective effect is to
-# 	"change the headers so that it uses 1x 1023-byte isochronous endpoint,
-# 	rather than 6x 128-byte endpoints to send the scope/logic analyzer data"
-# 	(from https://github.com/espotek-org/Labrador/issues/260)
+# 	In the build process for variant 0x01, the macro SINGLE_ENDPOINT_INTERFACE
+# 	is undefined, while in the build process for variant 0x02,
+# 	SINGLE_ENDPOINT_INTERFACE is defined.  In several regions of the source
+# 	code, the undefined/defined status of this macro determines which of two
+# 	possible code blocks are compiled into the firmware.  The collective effect
+# 	of having SINGLE_ENDPOINT_INTERFACE defined is to "change the headers so
+# 	that it uses 1x 1023-byte isochronous endpoint, rather than 6x 128-byte
+# 	endpoints to send the scope/logic analyzer data" (from
+# 	https://github.com/espotek-org/Labrador/issues/260)
 # *NOTE* : there is a commented-out line defining SINGLE_ENDPOINT_INTERFACE in
 # globals.h.  If it is uncommented, both 'make 0x01' and 'make 0x02' will
 # produce variant 0x02.  Further, 'make 0x01' will mislabel the firmware with


mi-hol:
1. @brentfpage Unfortunately I lack the technical background to review it but have a questions.

From [this and further comments](https://github.com/espotek-org/Labrador/issues/384#issuecomment-3678699179) I understand that a Windows (*_01.hex) and a Unix (*_02.hex) variant need to be build and uploaded to Desktop_Interface/resources/firmware

1. Are the 2 variants implemented in the Makefile ?
2. Do you happen to know if the hex files are uploaded to Desktop_Interface/resources/firmware automatically?

brentfpage:
@mi-hol 
> Are the 2 variants implemented in the Makefile ?

I added the following comment that addresses this to the end of the "Building the firmware" section of the wiki:

"Referring to the variant identifiers `_0{1,2}.hex` in [the firmware releases](https://github.com/espotek-org/Labrador/tree/master/Desktop_Interface/resources/firmware), if this line is commented out, the build produces an `01` variant, and otherwise the build produces an `02` variant."

Here, "line" means `#define SINGLE_ENDPOINT_INTERFACE` in `globals.h`.
> Do you happen to know if the hex files are uploaded to Desktop_Interface/resources/firmware automatically?

I'm leaning towards no because I don't see any mechanism in `.github/workflows` for this.

mi-hol:
> I added the following comment

I'd understand this as "no, the 2 variants are not build automatically"

Could this be automated?

brentfpage:
@mi-hol 
> Could this be automated?

Maybe using https://github.com/marketplace/actions/avr-make

EspoTek:
AFAIK `make CFLAGS+=SINGLE_ENDPOINT_INTERFACE` _should_ add a #define for SINGLE_ENDPOINT_INTERFACE and therefore build variant 02?

brentfpage:
In my limited experience I have used compiler flags like -D(thing to be #defined) for this purpose, e.g., `-DSINGLE_ENDPOINT_INTERFACE`

mi-hol:
> In my limited experience I have used compiler flags

@brentfpage are you inclined to give this a trail?

brentfpage:
Based off [this](https://stackoverflow.com/questions/2129391/append-to-gnu-make-variables-via-the-command-line), it seems like one way to do this is along the lines of 
- delete the line `#define SINGLE_ENDPOINT_INTERFACE` in globals.h
- change `CFLAGS=...` in the Makefile to `CFLAGS+=...`
- to compile a version with `SINGLE_ENDPOINT_INTERFACE` defined, use `CFLAGS=-DSINGLE_ENDPOINT_INTERFACE make all`
    - I've confirmed that this produces a version of the firmware equivalent to that produced by `make all` when `#define SINGLE_ENDPOINT_INTERFACE` is present in `globals.h`
- to compile a version with `SINGLE_ENDPOINT_INTERFACE` NOT defined, use `make all`

If automatic naming is also useful, instead of the above, the following block should be added to the Makefile
```
CFLAGS=...
ifdef SINGLE_ENDPOINT_INTERFACE
	CFLAGS += -DSINGLE_ENDPOINT_INTERFACE
	SUFFIX = 0x02
else
	SUFFIX = 0x01
endif
NAME=labrafirm_$(SUFFIX)
```
and `$(NAME)` should be substituted for `labrafirm_new` throughout.  Then, 
`make -f Makefile_mine all`
creates variant 0x01 versions of the firmware appropriately named labrafirm_0x01 and 
`make -f Makefile_mine SINGLE_ENDPOINT_INTERFACE=1 all`
creates variant 0x02 versions of the firmware appropriately named labrafirm_0x02.

mi-hol:
> automatic naming is also useful, instead of the above, the following block should be added to the Makefile

@brentfpage with my limited technical knowledge, this approach looks perfect :)

mi-hol:
and the reasoning for the 2 variants for firmware should be added as comments to make file and relevant source code file(s)

brentfpage:
Settled on `make 0x01` builds variant 1 and `make 0x02` builds variant 2 (with automatic naming).  @EspoTek I included in the makefile header your explanation from #260 of the variant 0x01 vs. 0x02 distinction.  Is there anything else I should add?

mi-hol:
@brentfpage thank you so much! Looks perfect to me!
@EspoTek I plan to merge this PR tomorrow, unless you have a concern.

PR #389, opened 2026-01-03
Simplify scroll zoom calculations

turboencabulator:
I'm finding the code and math behind the plot extent calculations hard to follow.  There are many casts to `(double)` which are mostly unnecessary, either the expression is already a double or it's an integer literal that we can replace with a floating-point literal.  There are several places where we multiply or divide by 100 and they often cancel out.  This PR cleans it up and removes many of the extra operations.  There should be no functional change.


EspoTek:
Ship it.  I think I wrote that function while still a student with no training at all in C++!

Issue #288, opened 2025-02-13
Trigger status obscures spectral display


tdunning:
The trigger status display obscures some of the spectral display. Turning off the trigger is a work-around, but not at all obvious because the trigger seems to have no effect on the spectral display

One possible action would be to always hide the trigger display when the spectrum is being shown.

PR #390, opened 2026-01-03
Continuing #367 "Triggering at sig. gen. freq" with a cleaner file history

brentfpage:
Added new trigger type options, Sig. Gen. Ch1 and Ch2 , that set the trigger frequency equal to the respective sig. gen. frequencies. In these modes, there's no need to adjust the trigger level to get the waveform steady on the screen.  The feature gets finicky above \~25 khz or so, especially when 'scope channel 2 is enabled. I've dug around trying to find a cause, but haven't had any luck yet. The issue only appears in a limited operating range (>~ 25 khz), is not all that common even in that range, and can usually be resolved by disabling/enabling 'scope channel 2, so imo isn't a huge demerit for the PR.  Also, #288 is fixed in this PR.

### DIFF
--- a/Desktop_Interface/genericusbdriver.cpp
+++ b/Desktop_Interface/genericusbdriver.cpp
@@ -253,6 +253,7 @@ void genericUsbDriver::sendFunctionGenData(functionGen::ChannelID channelID)
 			usbSendControl(0x40, 0xb2, timerPeriod, clkSetting, channelData.samples.size(), channelData.samples.data());
     }
 
+    this->sigGenFreqUpdated(channelID, clkSetting, timerPeriod, channelData.samples.size());
     return;
 
 }
--- a/Desktop_Interface/genericusbdriver.h
+++ b/Desktop_Interface/genericusbdriver.h
@@ -121,6 +121,7 @@ signals:
     void initialConnectComplete(void);
     void signalFirmwareFlash(void);
     void calibrateMe(void);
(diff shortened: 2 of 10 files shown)


EspoTek:
Just for what it's worth, the signal gen runs from a 1MSPS DAC, and at higher frequencies the waveform reduces in resolution to compensate.

At higher frequencies I'd expect quantisation noise by itself to get to something like 5% of the period of a waveform.

brentfpage:
I think I correctly handled 
> at higher frequencies the waveform reduces in resolution to compensate

but I should double check.

> At higher frequencies I'd expect quantisation noise by itself to get to something like 5% of the period of a waveform.

Can you please elaborate?  Say a single period of the waveform gets produced over the course of 2^10 clock cycles and a single 'scope sample is collected every 2^3 clock cycles.  If you then add a trigger point every 2^7 'scope samples, shouldn't that keep the waveform steady on the screen?

mi-hol:
> but I should double check.

@brentfpage are we ready to merge?

brentfpage:
> are we ready to merge? 

@mi-hol I'd prefer to spend a bit more time trying to improve the performance at >25 khz

mi-hol:
> I'd prefer to spend a bit more time trying to improve the performance at >25 khz

@brentfpage that is at best signaled to others via setting the PR to "draft" status by clicking on "convert to draft" link.

image

I've applied that for you. As soon as it's ready from your view just remove "draft" status by pressing 
"Ready for review" button
image

brentfpage:
@mi-hol Ah OK thanks

PR #393, opened 2026-01-05
flake.nix: add macOS (Darwin) support

randomizedcoder:
G'day Chris,

I just created the nixpkgs pull request, so it will be easy for people to see Labrador is supported by Nixpkgs.  Once merged it will also sort of make the Flake redundant.
https://github.com/NixOS/nixpkgs/pull/477199

However, I noticed that I would also improve the platform support adding Darwin to the flake.

- Add platform detection for Darwin vs Linux
- Handle different libdfuprog paths and extensions (.so/.dylib)
- Use appropriate library path env var (LD_LIBRARY_PATH/DYLD_LIBRARY_PATH)
- Handle binary name differences (labrador/Labrador)
- Install icons and desktop files only on Linux
- Use sysctl fallback for CPU count on macOS
- Update platforms to include darwin

I don't have a recent Apples devices, so this is untested.

Thanks,
Dave


EspoTek:
If it's not possible to test on macOS, I think we just keep support limited to Linux.

randomizedcoder:
Makes sense.  If we do find somebody with Darwin, then it would be great to test it!

mi-hol:
> find somebody with Darwin,

While this is "developer speak" I believe it means [any macOS or iOS version](https://en.wikipedia.org/wiki/Darwin_(operating_system)) should work for testing.
@EspoTek I though you'd own at least one device running macOS

mi-hol:
@randomizedcoder I have never heard of NixOS before and had just a little bit of reading.

Therefore I ask myself:
How will this package find its way into the official NixOS  repo so the search https://search.nixos.org/flakes?channel=25.11&query=Labrador will find it?

randomizedcoder:
If you are still using a non-deterministic operation system (e.g. Debian or Ubuntu), then it's definitely time to try NixOS

I submitted the pull request a while back
https://github.com/NixOS/nixpkgs/pull/477199

Looks like I need to fix the pull request

EspoTek:
Let us know when both PRs are ready to go and we can merge them together.

mi-hol:
@EspoTek as the PR has no negative impact on Labrador I'd suggest to merge it now.
Please raise any objections by 27Mar.

turboencabulator:
1. The latest merge needs to be redone, it still mentions libdfuprog binaries that are now gone.  It should be simpler now that it doesn't have to deal with the library handling differences.
2. `installPhase` has a lot of manual steps, is it possible to replace them with something like `make INSTALL_ROOT=$out install`?  (I'm not sure if `$out` is the correct thing to use there since it's the same as `PREFIX`.)
3. Are these scripts something that ultimately belongs in the NixOS/nixpkgs repo instead of here?  I'm not familiar with how it does things, but typically most projects don't tend to incorporate distro-specific packaging scripts (due to the effort involved by the distro to put their scripts everywhere, and again the effort involved if the distro needs to make a breaking change to those scripts).

PR #395, opened 2026-01-07
Status indicator and amplitude notice for the frequency response tool

brentfpage:
In the frequency response tool mode, if the input waveform has a p2p amplitude less than ~500 mV, the code typically finds the fit residuals to be too high and rejects the data.  Then, nothing gets plotted and it isn't obvious what is going wrong.  This PR adds:
- a status indicator for the tool that shows a green check mark when the least squares fitting succeeds and a red X when it finds unacceptable residuals
 - a notice recommending that the user increase the amplitude of Signal Gen CH1 if they select the tool with a p2p amplitude <500mV

### DIFF
--- a/Desktop_Interface/mainwindow.cpp
+++ b/Desktop_Interface/mainwindow.cpp
@@ -2791,6 +2791,8 @@ void MainWindow::on_actionFrequency_Spectrum_triggered(bool checked)
         ui->cursorHoriCheck->setChecked(ui->controller_iso->horiCursorEnabled1);
         ui->cursorVertCheck->setChecked(ui->controller_iso->vertCursorEnabled1);
         ui->controller_iso->retickXAxis();
+        ui->controller_iso->freqRespStatusMark->setVisible(false);
+        ui->controller_iso->freqRespStatusLabel->setVisible(false);
     }else{
         ui->cursorHoriCheck->setChecked(ui->controller_iso->horiCursorEnabled0);
         ui->cursorVertCheck->setChecked(ui->controller_iso->vertCursorEnabled0);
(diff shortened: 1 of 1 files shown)


mi-hol:
@mmehari are you ok to review it?

mmehari (approved):
Sorry for the late response and I would be glad to review it.

Indeed I have also experienced what you have witnessed and I have already captured it in the wiki page (https://github.com/espotek-org/Labrador/wiki/Frequency-Response-Calculation#special-note).

As to the improvements you came up with, I like the indicator and it is visually appealing to know the status of the frequency response at a given frequency.
I also like the recommendation to the user when the input voltage is less than 500mv. However, the recommendation pops up only when entering to the frequency response page. I am not sure if you made it intentionally, thinking that it might distract the user when ever a voltage below 500mv is selected, but I like the improvements you made.

brentfpage:
@mmehari Many thanks for reviewing.  To support first-time users, I think it is worthwhile to supplement the wiki note with the pop-up recommendation.  I did intend for it to only appear when the freq. response tool is entered and the input voltage is less than 500 mV p2p.  The main goal is to avoid having the tool selected and then appear inactive without explanation.

By the way, is there an option currently to view the 'scope traces with those least-squares fits overplotted, like shown in the wiki?  If not, I'd be interested in adding such a display and would be happy to implement it.

Wish I had this tool when trying to make a switched-capacitor filter using a 4000-series CMOS chip a couple years ago.  Exporting data to python was a major drag.  Was never able to get the filter to work quite right, and maybe it's prime time to return to it.

mmehari:
The square fitting plot on the top of the measurement data, shown in the wiki page, is made offline only for the Wiki page but if you want to make it part of the measurement plot, please do so. I will be curious how you want to put it together

PR #411, opened 2026-02-15
indenting #ifdef directives to level of surrounding code?

mi-hol:
also make conditional logic in preprocessor directives (i.e for PLATFORM_ definitions) better understandable  by indenting them to level of surrounding code.

Note: no functional changes!

### DIFF
--- a/Desktop_Interface/mainwindow.cpp
+++ b/Desktop_Interface/mainwindow.cpp
@@ -5,11 +5,11 @@
 #include "espospinbox.h"
 
 #if defined(PLATFORM_WINDOWS)
-#include "winusbdriver.h"
+    #include "winusbdriver.h"
 #elif defined(PLATFORM_ANDROID)
-#include "androidusbdriver.h"
+    #include "androidusbdriver.h"
 #else
-#include "unixusbdriver.h"
+    #include "unixusbdriver.h"
 #endif
 
 #include <algorithm>
@@ -23,10 +23,10 @@ namespace
    constexpr char kDocumentationUrl[] = "https://github.com/espotek-org/Labrador/wiki";
    constexpr char kPinoutUrl[] = "https://github.com/espotek-org/Labrador/wiki/Pinout";
    constexpr char kAboutString[] = "<h4>EspoTek Labrador</h4>"
-#ifdef GIT_HASH_SHORT
+    #ifdef GIT_HASH_SHORT
                                    "Continuous Release<br>"
                                    "Git hash: <a href='https://github.com/espotek-org/Labrador/commits/" QUOTE(GIT_HASH_SHORT) "'>" QUOTE(GIT_HASH_SHORT) "</a><br>"
-#endif
+    #endif
                                    "Website: <a href='http://espotek.com'>https://espotek.com</a><br>"
                                    "Contact email: <a href='mailto:admin@espotek.com'>admin@espotek.com</a>";
    constexpr char kOrganisationName[] = "EspoTek";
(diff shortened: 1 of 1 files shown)


mi-hol:
@turboencabulator @mmehari I'd appreciate a quick feedback if these changes make the code better readable from your view.
Note: this is just a proof of concept and not finalized yet!

mmehari:
I would prefer to keep it the way it was since there are IDEs which grey out the not used code and it appears as if it was written without the directives.

mi-hol:
> there are IDEs which grey out the not used code and it appears as if it was written without the directives.

I'm not following, would you perhaps have a screenshot of what you tried to explain?

mmehari:
Here is a screenshot from Qt Creator when viewing mainwindow.cpp file in the editor page.

image

Since I am working on a Linux machine, those directives related to Windows are grayed out and wouldn't bother if kept without being indented.

mi-hol:
> screenshot from Qt Creator

@mmehari thanks, that triggers new questions.
1. What version of QT Creator do you use?
2. How do the lines 58/following look in QT Creator before and after the change?

mmehari:
> > screenshot from Qt Creator
> 
> @mmehari thanks, that triggers new questions.
> 
> 1. What version of QT Creator do you use?

Qt Creator 18.0.2

> 2. How do the lines 58/following look in QT Creator before and after the change?

Before
image

After
image

mi-hol:
> Since I am working on a Linux machine, those directives related to Windows are grayed out and wouldn't bother if kept without being indented.

The screenshots leave me confused, because from my view the behaviour of QT Creator is the same for indented and non-indented directives.
But users of other editors benefit from the easier visiual destinction of conditional code identified by indents.
Do I miss something ?

mmehari:
Yes, the behaviour is the same since QT Creator only greys out the code that will not be compiled.

You didn't miss anything and I understand your point of view which at the end is a matter of preference. 

I am OK with your changes as well.

mi-hol:
> which at the end is a matter of preference

From my view it's much more than personal preference because it "can reap significant improvements in comprehension and productivity". Reasoning is well explained in https://thelinuxcode.com/why-you-should-use-column-indentation-to-improve-your-codes-readability/ 
Maybe that explains why I bring such topics up and invest my time there.

mmehari:
I think it is getting a bit out of context.

For the record, I am in favor of indentation and I am not saying otherwise. But specific to this situation, pre-processor directives are resolved before compilation and people tend to do the indentation at this level. Which is why IDE editors grey them out and the reason for my preference.

EspoTek:
I'm with @mmehari on this one too.

It seems strange to me to indent an #ifdef, since it doesn't compile to anything and isn't a part of the regular code.

Preprocessor stuff starting at column 0 seems to be the style that Atmel use too:
https://github.com/espotek-org/Labrador/blob/master/AVR_Code/USB_BULK_TEST/src/ASF/common/services/usb/udc/udc.c

mi-hol:
> Which is why IDE editors grey them out

unfortunately not all IDEs have that feature.
I'll make sure to mention this great feature of QTCreator in the developer doc as the preferred IDE.

I understand now the reasoning behind your preferences much better and will close this PR, after updating the developer doc

EspoTek:
I wouldn't say Qt Creator specifically is the preferred IDE, on macOS/Linux especially I'd say Cursor or VS Code are easier to work with.

mi-hol:
Does VS Code have the feature to "grey out the not used code and it appears as if it was written without the directives." 

On Windows VS Code I don't see such a feature by default
image

mmehari:
In Linux (ubuntu 24.04) VS Code, I see the following
image

mi-hol:
superseeded by #437

PR #412, opened 2026-02-16
Restore trailing double spaces in README.md

randomizedcoder:
The trailing double spaces are intentional Markdown syntax to  force line breaks. They were inadvertently removed in commit 3119205.


mi-hol:
@randomizedcoder Extremely fast fix, thank you so much!

PR #413, opened 2026-02-20
move librador isoCallback mutex positions

brentfpage:
This PR for librador changes the position of the mutex lock/unlock in usbcallhandler.cpp->`isoCallback(...)` such that the mutex stays locked for an entire USB transfer's worth of data (33 packets) instead of only a single packet.  For context, `isoCallback` gets called by libusb when a USB transfer is ready and proceeds to read in the new data to usbcallhandler.cpp.
Justification:
	Say a front-end gui is operating at 60 fps and so is requesting data from usbcallhandler.cpp every 17 ms.  If one of these requests comes when `buffer_read_write_mutex` is locked by `isoCallback`, then the request waits until the mutex gets unlocked.  If it gets unlocked after every packet, the request then accesses all data up to that most recent packet and finishes.
	The gui then won't request data again for another 17 ms.  However, at the time of the original request, all packets of the most recent USB transfer are available but simply haven't yet been added to usbcallhandler's buffers by `isoCallback`.  This process takes only something like a millisecond.  This PR's repositioning of the mutex lock/unlock in `isoCallback` trades this 1 ms delay for the 17 ms delay.

### DIFF
--- a/Librador_API/___librador/librador/usbcallhandler.cpp
+++ b/Librador_API/___librador/librador/usbcallhandler.cpp
@@ -53,10 +53,10 @@ o1buffer *internal_o1_buffer_750;
 static void LIBUSB_CALL isoCallback(struct libusb_transfer * transfer){
     //Thread mutex??
     //printf("Copy the data...\n");
+    buffer_read_write_mutex.lock();
     for(int i=0;i<transfer->num_iso_packets;i++){
         unsigned char *packetPointer = libusb_get_iso_packet_buffer_simple(transfer, i);
         //TODO: a switch statement here to handle all the modes.
-        buffer_read_write_mutex.lock();
         switch(deviceMode){
         case 0:
             internal_o1_buffer_375_CH1->addVector((char*) packetPointer, 375);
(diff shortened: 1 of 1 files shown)


EspoTek:
I'm a little bit confused by what this is doing.  Is it not just adding delay to the stream?  Why do we want to do this?

brentfpage:
It adds a small delay for packets already read in by `isoCallback` but removes a larger delay for packets of the most recent USB transfer that haven't yet been read in by `isoCallback`.

EspoTek:
I see what you mean.

Lets say the actual transfer group finishes at time t=1001ms, but isoController polls at time t=1000ms.
With your mutexes, it'll wait 1ms for the data to be ready then draw at time t=1001.  With my mutexes, it'll fetch nothing then try again at t=1017.  This could also cause occasional stutter as depending on the phase of the clocks this delay would drift between 0ms and 17ms over time.

Very subtle pickup!  I'll merge now.

Will bump up the framerate on the macos version to 60Hz too.

brentfpage:
Thanks!  I'll have to think more about:
> This could also cause occasional stutter as depending on the phase of the clocks this delay would drift between 0ms and 17ms over time.

Still trying to wrap by head around how these ~1-10 ms timing deltas have a perceptible impact.  Seems like it involves strobe effects in the waveform presentation.  For example, a 100 Hz waveform moves 10.1 periods in 0.101 seconds, but this looks like only 0.1 periods when viewed at a 10 fps frame rate.  This produces the illusion of a relatively slowly translating waveform, and changes to the frame rate can disrupt this illusion.

PR #418, opened 2026-02-23
Update mac.yml for workflow triggers and environment variables

mi-hol:
Refactor workflow triggers and clean up environment variable usage.

### DIFF
--- a/.github/workflows/mac.yml
+++ b/.github/workflows/mac.yml
@@ -4,13 +4,22 @@ permissions:
   contents: read
   pull-requests: write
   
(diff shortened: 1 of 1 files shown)


EspoTek:
Just for what it's worth we don't have a "main" branch (AFAIK), just "master"

mi-hol:
> we don't have a "main" branch (AFAIK), just "master"

the new standard name is "main" (ie in GH documentation), "master" is legacy and therefore I started to prepare for a migration to "main"

PR #428, opened 2026-03-08
Build and static-link libdfuprog

turboencabulator:
The current handling of libdfuprog is problematic:
* Requires building it for each platform/architecture and copying the library binary into this repo.  Bloats this repo and requires extra work to target new platforms.
* Distros would generally prefer to build libdfuprog from source instead of trusting a binary blob in this repo that is hard to audit.
* It's currently built as a dynamic lib, requiring extra effort to get the labrador executable to find it.

This deletes all copies of libdfuprog from this repo.  Instead we add dfu-programmer as a submodule, copy over the modifications to make it a library and work on Android, and fix a few bugs in the process:
* Rename `main.c` to `libdfuprog.c` and remove `main()` to avoid conflicts.
* The `commandLine` argument to `dfuprog_virtual_cmd()` is now const.
* Grow the `argv` array by 1.  Needs to be the number of spaces in the command + 2, assuming number of tokens <= number of spaces + 1, plus another for `argv[argc] = NULL`.
* Grow the `commandLine_copy` buffer by 1, needed for the terminating NUL.  Simplified using `strdup` instead of `malloc`+`strcpy`.
* Free both `argv` and `commandLine_copy` when done to avoid a leak.
* Loop until `strtok` returns NULL instead of assuming it will find exactly `argc` tokens (where `argc` was guessed from the number of spaces).
* Inline the Android-specific `dfu_device_init()` into `dfuprog_virtual_main()` so we don't have to modify `dfu.c` and `dfu.h`.

`libdfuprog.pro` builds a static lib.  The project isn't organized in a way that makes it easy for qmake to automatically build that separate project and link to it, so for now I have `libdfuprog.pri` which adds all of these files into the `Labrador` project/executable.


mi-hol:
Greatly appreciated 👍

mi-hol:
@EspoTek in light of https://github.com/espotek-org/Labrador/commit/4ab15552ff7b2abab1354a1eee41b7b46b2e5687, could you please review this PR?

From my view actively managing a dependency, as proposed here, seems a much better & future proof approach but I lack the technical expertise.

EspoTek:
> @EspoTek in light of [4ab1555](https://github.com/espotek-org/Labrador/commit/4ab15552ff7b2abab1354a1eee41b7b46b2e5687), could you please review this PR?
> 
> From my view actively managing a dependency, as proposed here, seems a much better & future proof approach but I lack the technical expertise.

Funny timing, eh?  This PR is the real long-term fix, mine was just a quick hack in response to a user email about it crashing on their Macbook.

EspoTek:
Yep, this looks great.  The one question I have:
Will adding dfu-programmer as a submodule cause their changes to be auto-merged into ours?

If so I think it would be best to fork it to something owned by espotek-org.  I've been burned in the past when upstream projects made breaking changes.

mi-hol:
> Will adding dfu-programmer as a submodule cause their changes to be auto-merged into ours?

https://github.blog/open-source/git/working-with-submodules/#advice-on-using-submodules-or-not  

recommends:
"Submodules require you to carefully balance consistency and convenience. The setup used _here(refers to example at https://github.blog/open-source/git/working-with-submodules)_  strongly prefers consistency, at the cost of a little convenience. **It’s generally best to have a project’s submodules locked to a specific SHA,** so all collaborators receive the same content. But this setup also makes it difficult for developers in the “parent” repository to contribute changes back to the submodule repository."

"SHA" in above means a specific commit from my understanding.
With that approach the answer to initial question would be a "No".

@turboencabulator is this correct understanding?

turboencabulator:
A submodule always points to a particular commit.  In this case it's pointing to the same commit where you forked it for libdfuprog.  It won't change unless we change it.

EspoTek:
Beautiful.  Is there anything else that needs doing on this, @turboencabulator, or are we good to ship?  Looks like my hotfix from the other day did not actually fully fix the issue.

turboencabulator:
I need to double-check some scripts to make sure they check out the submodule.  I'll take care of that and get it merged tomorrow.

EspoTek:
Too easy, cheers mate.  Feel free to just delete the new dylib I added when resolving the merge conflict.

EspoTek:
Nice work!  Running the release now

Issue #382, opened 2025-12-20
multiple Calibration dialogs open after USB disconnect/connect while executing calibration


mi-hol:
### PROBLEM DESCRIPTION
Calibration is not protected against multiple activation during a pending calibration

### REQUESTED INFORMATION

- [x] Verified with latest published [Labrador version](https://github.com/espotek-org/Labrador/releases)
- [x] Searched the problem in the [Troubleshooting Guide](https://github.com/espotek-org/Labrador/wiki/Troubleshooting-Guide)
- [x] Searched the problem in the [docs](https://github.com/espotek-org/Labrador/wiki)
- [x] Searched the problem in [issues](https://github.com/espotek-org/Labrador/issues)
- [x] Searched the problem in [discussions](https://github.com/espotek-org/Labrador/discussions)

- Hardware the software is running on: "System Type: x64-based PC"
- Operating System (OS) Build Version: "OS Name: "Microsoft Windows 11 Home" & "OS Version: 25H2 (Build: 26200.7462)" & "OS Configuration:      Standalone Workstation")
- Labrador Git hash:  572b5577ae30ef21db59a318a3bd30c91987c1aa

### TO REPRODUCE
_Steps to reproduce the behavior:_
i.e.:
1. Go to 'Oscillospe/Calibrate'
3. touch USB Shield, triggers a short disconnect of USB connection and instant reconnection
4. See a new calibration dialog

### EXPECTED BEHAVIOUR
Only a single calibration dialog while calibration was started but not completed yet

### SCREENSHOTS

<img width="890" height="401" alt="Image" src="https://github.com/user-attachments/assets/4559557d-a161-46a1-9f02-92f23a6c99af" />

3 dialogs are open but only one visible due to stacked windows of same size
<img width="1146" height="658" alt="Image" src="https://github.com/user-attachments/assets/3453dc2a-e479-42b8-9bf8-e4508c9aa588" />

### ADDITIONAL CONTEXT
was fixed for Android in https://github.com/espotek-org/Labrador/pull/376
Likely this needs the same approach for all platforms

brentfpage:
Edit: I read the title of the post again and now better understand.  This could very well happen on mac as well, and I'll look into fixing it.

This might have been caused by the changes in #376.  I tested #376 on mac and android, but that obviously wasn't enough.  Just to make sure, the "Please connect ... to the Power Supply..." dialog box comes up in front of the previous "Please connect ... the USB connector..." box when the USB shield is touched?

[...]

Issue #333, opened 2025-05-22
2 race conditions in init phase leading to invisible USB related modal dialog boxes perceived as hangs


AntoineBlaisENAC:
Hello,

I am running the application on Debian GNU/Linux 12 (bookworm) after compilation from source.

Here is the content of the file labrador.log after running the command `./labrador 2>labrador.log`:

[labrador.log](https://github.com/user-attachments/files/20394968/labrador.log)
It seems to be a problem when using the libusb library.

Any help would be greatly appreciated,

Antoine

EspoTek:
Does the same problem occur when you run the AppImage?

AntoineBlaisENAC:
Yes unfortunately, exactly the same behavior with Labrador-82ea67c-x86_64.AppImage.
But maybe I have found a workaround: run the application, then plug the card. It seems to work normally, even if quitting with a core dump.

chmielar:
I'm experiencing the same issue. Thank you for the tip, @AntoineBlaisENAC; it worked for me as well. I tested both the .deb and .AppImage versions, and I got the same result. I'm running Kubuntu 25.04, and the problem started after I recently upgraded from 24.10. Please let me know how I can help. I'm willing to run tests and provide any details you may need.

EspoTek:
Thanks for sharing, guys.

So to confirm:
 - The issue occurs with both the source and the AppImage version
 - It only segfaults if the board is connected before the app is launched.
 - It only happens on newer versions of libusb (it was fine with whatever was included in *ubuntu 24.10, but not 25.04)

turboencabulator:
I don't think the AppImage bundles libusb:
```
Skipping deployment of blacklisted library /lib/x86_64-linux-gnu/libusb-1.0.so.0
```

EspoTek:
That'd do it.  Mind you, we should probably be using the system libusb anyway, right?

chmielar:
@turboencabulator I'm not sure if it's relevant, but I also installed the application via .deb and encountered the same issue. Before upgrade to Kubuntu 25.04 it was good.

turboencabulator:
In all three cases (building from source, AppImage, .deb) it should be using the system libusb.  Which version of libusb do you have?

chmielar:
```
dpkg -l | grep libusb
ii  libgusb2:amd64                                           0.4.9-1                                    amd64        GLib wrapper around libusb1
[...]

PR #431, opened 2026-03-11
Windows race condition fix

freshLobster:
this is the same patch I did for issue #333, but applied to the windows specific code. It builds, it runs fine, tests fine, ~~logs look good save for:
Frame phase error of 3
in the logs which just tells me that there's some minor isochronous endpoint frame alignment that needs to be adjusted.
That should have it's own separate issue and PR though. this is for #333 and #382~~
Edit: fixed the alignment/ordering


mi-hol:
I did a basic test with Multimeter (mode 5) on Windows 11 and it works.

@turboencabulator would you perhaps have the time for a code review?

EspoTek:
"Frame phase error of 3" means that the data might be copying across into the buffer in the wrong order.

Can you see any discontinuities when feeding back a "weird" waveform like a 792Hz sine wave into the scope CH1?

freshLobster:
> "Frame phase error of 3" means that the data might be copying across into the buffer in the wrong order.
> 
> Can you see any discontinuities when feeding back a "weird" waveform like a 792Hz sine wave into the scope CH1?

I haven't tested that yet. I can put this PR as a draft and open it again when the frame phase error is corrected? 

Otherwise I can test it when I have time or someone else can feel free to check it out.

mi-hol:
> put this PR as a draft

just set it to DRAFT status

turboencabulator (commented):
I've only got a few minor suggestions.  This isn't a very thorough review, I'm not familiar with the Windows stuff and I don't have easy access to a Windows machine.

freshLobster:
I could use some help on fixing this. I dont know what happened, but my local build got all messed up and I cant get it back to the state where it worked properly.

EspoTek:
What's going on?  Does it chuck up build errors?

freshLobster:
@EspoTek no, it builds fine but for some reason after I got everything worked out a rebuild after a minor adjustment broke the whole thing and I havent been able to get back to a commit that has the working functionality I had before. one of those gremlins messing with my code.

EspoTek:
That's a pain.  I noticed you were rebasing against master a lot.  In general, I'd advise against this, and instead just do one cleanup at the end if strictly necessary.  Can help prevent instances like this.

Also `git reflog` is great for rewinding to an exact commit you know worked!

EspoTek:
Oh no, what happened!?

freshLobster:
@EspoTek I decided instead of untangling my work on this issue I will start over on the windows fix and use a more structured approach that is in line with the same process the unix patch used. Hopefully this will solve the calibration dialog issue, but itf it doesnt I will address that in it's own PR.

PR #434, opened 2026-03-17
Zoom gesture changes

brentfpage:
~~This PR changes the zoom gestures so that they're possibly more intuitive when a trackpad is used.  Specifically, placing the cursor on the l.h.s. of the screen and swiping left moves the graph leftward below the edge, and similarly for the r.h.s as well as the top and the bottom.  Placing the cursor in the middle 1/2 of the screen and swiping left/right zooms the screen out/in while keeping the data value of the exact center of the screen constant, and similarly for swiping down/up.
One drawback is that these new gestures no longer keep the x/y value hovered by the mouse constant during the zoom, which the current gestures do.  Such behavior is especially appealing for pinch gestures on android, for which the effective mouse position is the center of the user's fingers.  So, this PR includes preprocessor directives such that the PR has no effect on the android app.~~

### DIFF
--- a/Desktop_Interface/isodriver.cpp
+++ b/Desktop_Interface/isodriver.cpp
@@ -389,11 +389,12 @@ void DisplayControl::setRespAndSpecRanges(QWheelEvent* event, QCustomPlot* axes,
         range.lower += scale * offset;
         if (range.upper > 90.0) range.upper = 90.0;
         if (range.lower < -90.0) range.lower = -90.0;
-        topRange = range.upper;
-        botRange = range.lower;
-
-        topRangeUpdated(topRange);
-        botRangeUpdated(botRange);
+        if( (range.upper - range.lower) > 1.e-4) {
+            topRange = range.upper;
+            botRange = range.lower;
+            topRangeUpdated(topRange);
+            botRangeUpdated(botRange);
+        }
     } else {
         QCPRange range = axes->xAxis->range();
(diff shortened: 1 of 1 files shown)


brentfpage:
I've soured on most of the changes originally tried in this PR, but I think the remaining ones are good.  They 
1.) prevent the rate of the X/time zoom from slowing as the cursor position approaches the lhs of the screen
2.) simplify the clamping of the X/time zoom to (min(delay) = 0s , max(window) = 10s).

EspoTek:
Semi-tangential, but now that you have write access to the repo are you able to make your branches under espotek-org/Labrador rather than your personal fork?  That way it's easier for other maintainers to do things like run Actions on your branches.

EspoTek:
Also semi-tangentially related to this, but I realised that if you zoom in too far you can get "stuck" and can't zoom back out again.  This also occurs in master!

brentfpage:
> now that you have write access to the repo are you able to make your branches under espotek-org/Labrador rather than your personal fork?   That way it's easier for other maintainers to do things like run Actions on your branches.

That makes sense – I'll do that in the future
> Also semi-tangentially related to this, but I realised that if you zoom in too far you can get "stuck" and can't zoom back out again

The newest commit fixes this

EspoTek:
Happy to give my "ship it" stamp to this one.  :)

PR #442, opened 2026-04-17
break out of dfu_make_idle retry loop upon success

brentfpage:
This PR fixes a bug in the Android-specific stage of firmware flashing that runs `dfu_make_idle`.  Both pre-PR and in the PR, a loop is used to allow for retries in case the `dfu_make_idle` call fails.  However, in the pre-PR code, the loop isn't broken when `dfu_make_idle` is successful.  In the PR, the bool `made_idle` is used to break the loop upon success.  It seems that standard bools aren't available for the adopted C dialect, so I used the `dfu_bool` type from `dfu-programmer`.

I figure I'll also take this opportunity to document that the first call to `dfu_make_idle` fails with `status->bStatus: errSTALLEDPKT (0x0f)` and `status->bState: dfuERROR`, but the second try goes through just fine, flashing proceeds, and the firmware gets successfully uploaded.

### DIFF
--- a/libdfuprog/libdfuprog.c
+++ b/libdfuprog/libdfuprog.c
@@ -190,9 +190,11 @@ int dfuprog_virtual_main(int argc, char **argv)
     dfu_device.interface = interface;
 
     int retries = 6;
-    while (retries > 0) {
+    dfu_bool made_idle = false;
+    while (retries > 0 && made_idle == false) {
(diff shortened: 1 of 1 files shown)


EspoTek:
I have no idea what this means but I'll take your word for it, haha.  Ship it!

brentfpage:
Hahha, I promise it's beneficial!

PR #443, opened 2026-04-18
Android imgui app

brentfpage:
This PR
- creates a new folder Android_App in the repo root that contains the source files for the imgui android app
- adds submodule dependencies to Android_App/app/src/main/cpp/deps.  The submodules are forks of SDL, imgui, and implot that I created when working on the app
- changes android.yml so that it builds the new app 
- adds a CMakeLists.txt to libdfuprog.  This is required because the current build system used for the app is CMake (called by gradle).  I'm not sure how easy it would be to get away from CMake because it and ndk-build appear to be the only build systems supported by Android Studio. (https://developer.android.com/studio/projects/add-native-code).

Subsequent PRs:
1. Move the android-specific files in Desktop_Interface to Android_App/legacy .  I'll also remove the code wrapped by `#ifdef PLATFORM_ANDROID` in Desktop_Interface.
2. Move the files in repo_root/Librador_API/___librador/librador to a legacy directory and replace them with the files from this PR's Android_App/app/src/main/cpp/deps/librador.


EspoTek:
Looks good, ship it!

brentfpage:
Merged – the "Build Android" action now builds the new imgui-based app!

PR #445, opened 2026-07-07
Final Qt desktop release: macOS moves to firmware 12.3

EspoTek:
The only change going into master: the Qt app's **macOS build moves to firmware 0x000C variant 03 ("12.3")**, streaming from the AIO firmware's bulk interface — opening a full-speed iso pipe kernel-panics macOS Tahoe (IOUSBHostFamily NULL-dereference on the missing SuperSpeed companion descriptor). Verified live on macOS hardware: firmware match, bulk alt-setting selected, clean checksummed frames. `labrafirm_000c_03.hex` is bundled into all packages via the existing `labrafirm*` glob; the mac app auto-flashes it.

Windows / Linux / Raspberry Pi are untouched: still firmware 0x0007 variants 01/02, same transports, same installer contents.

Note: the `Build Android` check is expected to be red — it fails at gradle configure on CI (`keystore.properties` is a local-only file) and has failed on every master push since 2026-06-20. Pre-existing, unrelated to this PR, and deliberately not addressed here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


EspoTek:
good bot

